PR #91, opened 2026-03-30
groovy/ruby/smalltalk inspired dsl with scope

paul-hammant:
Per https://paulhammant.com/2024/02/14/that-ruby-and-groovy-language-feature/ the same "DSL with scope" feature for Aether


paul-hammant:
Versus **Gradle's build grammar** - specifically the kotlin one which is compiled inlike the classic one in groovy that's interepreted, the closure flavor of aether lacks type-safe builders. This is where the compiler would know what methods are valid inside each block. Aether doesn't have this as any function with `_ctx` gets injected with no scope restriction.   

An imagined build tech using Aether as it grammar for better or for worse might need that and maps/sets. Dags and task scheduling, file/process stuff and even plugins is just code via import.

**Nix-lang**

Not a Nix-lang replacement unless lazy evaluation is added , which would be a fundamental  change to Aether's execution model. Otherwise there's high match to that dsl. Well, unless I'm missing something.

PR #100, opened 2026-04-02
Lazy evaluation implementation with tests

paul-hammant:
I think this honors the constraints of https://github.com/nicolasmd87/aether/issues/94


paul-hammant:
Comments acted on. 
                            
That said, the thunk_free ownership decision is worth explaining: the thunk copies the _AeClosure struct by value, so both the closure variable and the thunk hold the same env pointer. The closure variable's auto-defer frees it at scope exit.  If thunk_free also freed it, it's a double-free. The thunk borrows, the closure owns.

PR #101, opened 2026-04-02
heredocs is a marvellous simplification of multi-line strings

paul-hammant:
Implementation of <<HEREDOCS for Aether with tests


paul-hammant:
feedback acted on :)

PR #104, opened 2026-04-04
Add runtime containment sandbox. Also six language host modules using the same

paul-hammant:
Sandbo^CDSL with deny-by-default grants, runtime enforcement via LD_PRELOAD libc interception, and embedded language host modules for Python, Lua, JS (Duktape), Perl, Ruby, and Java. No deny() as yet.

The contained code uses normal stdlib calls and has no idea it's sandboxed. Nested sandboxes can only narrow, never escalate. There is a stderr and to-log-file option for denied reporting, Or none at all.

Features:
 - grant_tcp, grant_fs_read, grant_fs_write, grant_exec, grant_env, grant_all
 - Glob patterns: prefix (/etc/*), suffix (*.example.com), exact, wildcard
 - spawn_sandboxed: cross-process via POSIX shared memory + LD_PRELOAD
  - Shared map: token-guarded string:string data exchange with frozen inputs
 - Intercepts: open, connect, getenv, execve, dlopen, mmap, mprotect, fork, bind
 - Six std/host/ modules with aether_map_get/put bindings
 - Denial logging to file (default), stderr, or silent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,5 +84,3 @@ zig_minimal
 .DS_Store
 Thumbs.db
 desktop.ini
-TODO_paul.md
-real-build-system.md


nicolasmd87:
Hey Paul, want to take this as one PR, no split needed. It's a great functionality all around, and it's very well implemented.

There are a few things that will break CI on macOS and Windows:

- `aether_spawn_sandboxed.c `and `aether_shared_map.c` are added to RUNTIME_SRC unconditionally. They include `<sys/mman.h>, <sys/wait.h>, <unistd.h>`; won't compile on Windows. Need `#ifdef __linux__` guards around the implementation with stubs that return error codes on other platforms, or gate them out of RUNTIME_SRC on non-Linux.

- `DAETHER_HAS_SANDBOX` is added to CFLAGS unconditionally. The stdlib checks are cross-platform and fine to always enable, but spawn_sandboxed needs the POSIX runtime which isn't available everywhere. Either split the flag (`AETHER_HAS_SANDBOX` for checks, `AETHER_HAS_SANDBOX_SPAWN` for the POSIX stuff) or make spawn_sandboxed a no-op on non-Linux.

- `make stdlib` tries to build `libaether_sandbox.so` with `-shared -fPIC -ldl -lrt` —> Linux-only. The `|| true` hides it, but it should be guarded with a platform check so it's not attempted on macOS/Windows at all.

- `spawn_sandboxed` is registered in the typechecker on all platforms, but the implementation only exists on Linux. A macOS/Windows user calling it gets a linker error instead of a clean message.

One design note for the future: the sandbox is process-global (one checker, one context stack). Works great now, but since Aether is an actor language with actors running on different threads, concurrent actors setting up different sandboxes would step on each other. Not a blocker, just worth keeping in mind. The natural next step would be the per-actor permission state.

The `std/host/*` modules (Python, Lua, etc.) —> are those meant to be compiled as part of the standard build? If so, they'd need the respective `-dev` packages. If they're reference implementations, maybe move them to a `contrib/` or `examples/host/` directory so they don't affect the CI build matrix (or an external dep used with `ae add xxx`).

paul-hammant:
I think I have enough now for a build system - diff repo of course

nicolasmd87:
this is great

PR #105, opened 2026-04-04
add named arguments, select() platform conditional, list literal tests

paul-hammant:
Three parser/codegen features that move Aether toward build DSL capability:                                                                                                            
   
  - **Named arguments**: `func(name: "alice", count: 3)` — uses colon syntax, consistent with Aether's `param: type` definition style. Names are documentation at the call site; C       
  receives positional args. Positional and named can be mixed.    
                                                                                                                                                                                         
  - **`select()` platform conditional**: Compile-time platform selection via named args. Emits `#ifdef _WIN32` / `__APPLE__` / `__linux__` chain in generated C. Supports `other:`       
  fallback. Integer values work fully; string values have a known type inference gap with `println` (use interpolation).
                                                                                                                                                                                         
  - **List literal tests**: Confirmed `[1, 2, 3]` and `["alice", "bob"]` array literal syntax already works — was in the parser/codegen but untested. Now has 11 assertions covering int 
  arrays, string arrays, computed expressions, and arithmetic on elements.
                                                                                                                                                                                         
## Motivation                                                   

  These are prerequisites for Starlark-competitive build DSLs (something I'm building up to):                                                                                                                           
  ```aether
  java_module("myapp") {                                                                                                                                                                 
      dep("components/model")                                                                                                                                                            
      compile()
      test() { surefire() }                                                                                                                                                              
  }                                                                                                                                                                                      
   
  flags = select(linux: "-lpthread", windows: "-lws2_32", macos: "-framework Security")                                                                                                  
  port = select(linux: 8080, windows: 80, other: 3000)            
                                                                                                                                                                                         
Test plan                                                       
                                                                                                                                                                                         
  - test_named_args.ae — 9 assertions: all positional, all named, mixed, expressions, strings                                                                                            
  - test_list_literals.ae — 11 assertions: int/string arrays, computed, arithmetic, single element
  - test_select.ae — 5 assertions: int select, arithmetic on result, fallback, conditional, multicore check  

CI passes, CHANGLEOG has new lines.

### DIFF
--- a/compiler/codegen/codegen_expr.c
+++ b/compiler/codegen/codegen_expr.c
@@ -856,6 +856,16 @@ void generate_expression(CodeGenerator* gen, ASTNode* expr) {
                                 default_val = arg->children[0];
                         }
                     }
+                    // Validate: every platform must have a value or other: must be set
+                    if (!default_val) {
+                        if (!linux_val || !windows_val || !macos_val) {
+                            fprintf(stderr,
+                                "error: select() at line %d: missing platform without 'other:' fallback.\n"
+                                "  Provide all platforms (linux:, windows:, macos:) or add other: for the default.\n",
+                                expr->line);
+                            // Still emit code so compilation continues and shows all errors
+                        }
+                    }
                     // Emit #ifdef chain
                     fprintf(gen->output, "\n#ifdef _WIN32\n");
                     if (windows_val) {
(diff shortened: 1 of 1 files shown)


nicolasmd87:
Nice changes.
select() falls back to 0 for missing platforms. If you write select(linux: 8080) without windows/macos, those platforms get 0. Should this be a compile error instead?
Can you add some more docs around the new functionality?

PR #120, opened 2026-04-07
fix module function return type inference across module boundaries

paul-hammant:
When calling a module function via namespace syntax (e.g., `result = mymod.greet("world")`), the type inferrer in `infer_type()` used `lookup_symbol()` which couldn't resolve dotted names. This caused all module function return values to default to `int`, breaking `string` and `ptr` returns.

Fix: use `lookup_qualified_symbol()` which correctly resolves `mymod.greet` → `mymod_greet` via the imported namespace registry. Added guards for `TYPE_VOID` and `TYPE_UNKNOWN` to avoid regressing pure-Aether functions whose return types haven't been fully inferred.

Also increased MAX_MODULE_TOKENS (2000 → 20000) and MAX_TOKENS (10000 → 50000) — large modules silently truncated at the old limits.

Regression test: tests/integration/module_return_types/ (6 cases covering string, ptr, int, nested, and conditional returns).


nicolasmd87:
Hey Paul, saw the repo, really cool use of the language. Seeing Aether used in the wild for the first time is invaluable.

PR #130, opened 2026-04-09
Add defer trailing-block functions, unqualified selective imports, and module system fixes    

paul-hammant:
## Summary                                                                                                                                                                             
                                                                                                                                                                                         
  - **Defer functions** (`defer` keyword): "configure then execute" DSL pattern where the
    trailing block fills a config map before the function runs. Supports custom factories                                                                                                
    via `with` clause.                                                                   
  - **Unqualified selective imports**: `import build (release, lint)` lets you call `release()`                                                                                          
    without the `build.` prefix — enables clean DSL blocks in trailing-block contexts.                                                                                                   
  - **Module system fixes**: defer functions now merge correctly from modules, `_ctx` injection                                                                                          
    works for module-qualified calls, duplicate imports no longer produce broken parameter names.                                                                                        
                                                                                                                                                                                         
## Motivation                                                                                                                                                                          
                                                                                                                                                                                         
These features were driven by [aetherBuild](https://github.com/paul-hammant/aetherBuild), which uses defer trailing blocks as a configuration DSL for build SDK functions (javac, kotlinc, cargo, etc). The unqualified import lets `.build.ae` files read cleanly:      
                                                                                                                                                                                         
``` 
  import build (release, lint, werror)                                                                                                                                                   
                                                                                                                                                                                         
  build.javac(b) {                                                
      release("21")
      lint("all")  
      werror()   
  }           
```
   
## `with` clause                                                                                                                                                                       
   
The default config object for a defer block is a `map_new()`, but `with`  lets SDK authors choose any zero-arg factory. `defer func(...) with list_new { ... }` gives the block a list instead of a map. `with my_builder_new` gives it a custom struct/builder. This means defer blocks aren't tied to key-value configuration - they can accumulate ordered items, build ASTs, populate domain-specific objects, or anything else a factory can produce.          

## Tests worth talking about:
                                                                                                                                                                                        
  - tests/syntax/test_closure_defer.ae — basic defer pattern
  - tests/syntax/test_closure_defer_edges.ae — edge cases                                                                                                                                
  - tests/syntax/test_closure_defer_factory.ae — custom factory via with                                                                                                                 
  - tests/syntax/test_selective_import_unqualified.ae — short-name imports
  - End-to-end: aetherBuild compiles and runs 18 compile + 17 test targets in google-monorepo-sim  (yet another repoof mine)


paul-hammant:
An example of the build grammar that's useing defer under the hood

https://github.com/paul-hammant/google-monorepo-sim/blob/trunk/rust/components/vowelbase/.build.ae

And this one doesn't have deferrable properties in the `javac` block (no `{ .. }` following `javac(..)` ) but does depend on the rust module above:

vhttps://github.com/paul-hammant/google-monorepo-sim/blob/trunk/java/components/vowelbase/.build.ae

PR #133, opened 2026-04-09
Feature/lib dir option

paul-hammant:
AetherBuild needs to have libs in .aeb/libs/ rather than default and this faciltates that


nicolasmd87:
Nice
A couple of questions:

1. **Global state** — `module_set_lib_dir()` looks like it sets module-resolver state. Is it reset between compilations? Not critical now, but worth noting if we ever expose the compiler as a library.

2. **Env var** — any reason not to also accept `AETHER_LIB_DIR` as an env var? It would let tools set it once in their shell rather than passing `--lib` on every invocation. Optional, up to you.

paul-hammant:
`ae` in a Language Server Protocol use would need to not have a global lib folder as global state - is that the rationale?  A forawrd looking use, and not just classic compile-and-exit tool. Cool. And yes to the env var, sure.

PR #137, opened 2026-04-12
fix: free intermediate strings on reassignment in loops...

paul-hammant:


nicolasmd87:
nice, I guess this was the leak?

PR #140, opened 2026-04-13
feat: reactor keep-alive mode for epoll HTTP benchmark (+5x throughput)

arielg81:
## Summary

- Add `--reactor` mode to `benchmarks/http/bench_epoll_http.c`
- Non-blocking keep-alive via scheduler I/O poller re-registration
- Each worker handles one request per `step()` invocation then re-registers the client fd — no scheduler thread ever blocks waiting for the next keep-alive request

## How it works

The current `--keepalive` mode blocks the scheduler thread in `recv()` between requests:

```
Worker (blocks scheduler thread)
  recv() ← waits up to 5ms for next request
  respond()
  recv() ← waits again ...
```

The reactor mode processes one request and yields immediately:

```
Worker step() invocation:
  recv()    ← data guaranteed (poller fired) or MSG_DONTWAIT
  respond()
  scheduler_io_register(my_core, fd, self, AETHER_IO_READ)
  return    ← thread is free immediately

Scheduler poller (no thread blocked):
  data arrives → MSG_IO_READY → worker step() called again
```

This reuses the `scheduler_io_register` / `MSG_IO_READY` / `IoReadyMessage` infrastructure added in #71. No runtime or stdlib changes required.

## Benchmark results

12-core machine, `wrk`, 8s duration:

| Mode | 100 connections | 1000 connections |
|------|----------------|------------------|
| `--keepalive` (blocking) | 29,773 req/s | 44,858 req/s |
| `--reactor` | **141,490 req/s** | **264,002 req/s** |

**4.8x improvement at 100c, 5.9x at 1000c.**

The blocking mode has ~311K socket read errors at 1000c (connections queued behind blocked workers). The reactor has zero errors.

## Changes

Single file: `benchmarks/http/bench_epoll_http.c`

- `worker_step_reactor()`: handles both `MSG_HTTP_CONNECTION` (new connection) and `MSG_IO_READY` (next keep-alive request). Sets fd non-blocking on first touch. Re-registers with `scheduler_io_register()` after each response. Falls back to `close()` if registration fails (cooperative/WASM mode).
- `--reactor` flag in `main()`: selects the new step function
- `#include <fcntl.h>` for `O_NONBLOCK`

## Test plan

- [x] Compiles with `-Wall -Wextra` (zero warnings)
- [x] All 166 existing tests pass (`make test`)
- [x] `--keepalive` and `--close` modes unaffected
- [x] Reactor stable at 100, 1000 connections (zero timeouts, zero errors)

### DIFF
--- a/benchmarks/http/bench_epoll_http.c
+++ b/benchmarks/http/bench_epoll_http.c
@@ -1,11 +1,14 @@
 /**
  * Benchmark: epoll-driven actor HTTP server with adaptive keep-alive
  *
- * Two worker modes:
- *   - Low concurrency:  keep-alive in worker loop (max throughput)
- *   - High concurrency: Connection: close (max concurrency)
+ * Three worker modes:
+ *   - keep-alive (default): blocking recv loop, max throughput at low concurrency
+ *   - close:                one request per connection, scales to high concurrency
+ *   - reactor:              non-blocking, re-registers fd with scheduler I/O poller
+ *                           after each response — handles 10K+ concurrent connections
+ *                           without blocking any scheduler thread
  *
- * The mode is controlled by a command-line flag: --keepalive or --close
+ * The mode is controlled by a command-line flag: --keepalive, --close, or --reactor
  * Default: keep-alive
  */
 #include <stdio.h>
@@ -17,6 +20,7 @@
 #include <sys/socket.h>
 #include <errno.h>
 #include <netinet/in.h>
+#include <fcntl.h>
 #include "../../std/net/aether_http_server.h"
 #include "../../runtime/scheduler/multicore_scheduler.h"
 #include "../../runtime/actors/actor_state_machine.h"
@@ -28,6 +32,7 @@
 
 static atomic_int total_requests = 0;
 static int use_keepalive = 1;  // default: keep-alive
+static int use_reactor = 0;
 
 // ---------------------------------------------------------------------------
 // Worker pool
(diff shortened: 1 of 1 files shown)


arielg81:
## Beyond the benchmark: reactor I/O as a language primitive

This PR demonstrates the reactor pattern working at the C level, but the infrastructure it builds on (`scheduler_io_register`, `MSG_IO_READY`, `IoReadyMessage`) is already part of the runtime. That opens the door to exposing this as a first-class language feature.

### The problem today

An Aether actor that handles keep-alive HTTP connections has to block a scheduler thread waiting for the next request. This caps concurrency at `num_cores * workers_per_core` simultaneous connections.

### What a language-level reactor would look like

```aether
actor HttpWorker {
  receive Connection(fd: Int) {
    let request = http_recv(fd)
    http_send(fd, build_response(request))
    await_io(fd)  // suspends this actor without blocking any thread
  }

  receive IoReady(fd: Int) {
    let request = http_recv(fd)
    http_send(fd, build_response(request))
    await_io(fd)
  }
}
```

`await_io(fd)` would call `scheduler_io_register` under the hood — the fd goes into the per-core epoll/kqueue/poll instance, the actor returns immediately, and the scheduler delivers `MSG_IO_READY` when data arrives. Between those two moments the actor consumes zero CPU and zero stack.

### Why this fits Aether's actor model

This is conceptually identical to how Erlang/Elixir processes suspend on `receive` without blocking OS threads. The difference is that Aether already has the scheduler infrastructure to do it — it just isn't exposed to the language yet.

The cost of keeping a connection alive moves from a **blocked scheduler thread** (expensive: stack, context switches) to a **kernel poll entry** (cheap: ~90 bytes in the epoll table). The same worker pool that handles 96 concurrent connections today could handle tens of thousands.

### Minimal implementation path

The compiler doesn't need to change for a first version. A stdlib binding would be enough:

```c
// std/net/module.ae — new extern
extern fn io_await(fd: Int) -> Void
```

```c
// std/net/aether_http_server.c — implementation
void aether_io_await(int fd) {
    // get current actor and core from TLS
    ActorBase* actor = g_sync_step_actor;
    int core = current_core_id;
    scheduler_io_register(core, fd, actor, AETHER_IO_READ);
}
```

The actor then handles `MSG_IO_READY` (type 300) in a `receive` block like any other message.

Happy to implement this if it's a direction you'd like to explore.

PR #147, opened 2026-04-14
Add hide and seal except scope directives (E0304)

paul-hammant:
## Summary

Two new scope-level directives that let a block decline to see selected names from its enclosing lexical scopes:

```aether
hide secret_token, db_handle           // blacklist specific outer names
seal except req, res, response_write   // whitelist — only these reach in
```

Both directives are compile-time hygiene — pure name-resolution denial at the boundary between a scope and its parent chain. No runtime overhead. No codegen change. Implemented entirely in compiler/analysis/typechecker.c.

This is the language-level expression of capability containment as a first-class concept: dependency injection delivers the names a block should reach, hide / seal except denies the rest. Without the denial side, DI is convention; with it, the dependency surface is enforced at the call site and auditable in one glance.

Background: [Principles of Containment](https://paulhammant.com/2016/12/14/principles-of-containment/)
(2016). The full design rationale, edge cases, and reference card live in docs/hide-and-seal.md. Then patronage of inversion of control in programming languages [in the years before before Dependency Injection was coined](https://paulhammant.com/blog/000128).

### Semantics in one paragraph

Scope-level (position within the block doesn't matter), reads AND writes both blocked, propagates to all nested blocks, does NOT reach through call boundaries (a visible function from an outer scope can still use the hidden name via its own lexical chain — hide is name resolution, not transitive information flow), local bindings inside
the hiding block are always visible, cannot redeclare a hidden name in the same scope. New error code E0304.

### What it looks like

A request handler that declares its dependency surface in one line:

```
handler(req, res) {
    seal except req, res, inventory, response_write, response_status
    // …everything else that's lexically in scope is now invisible.
    // The compiler enforces that this handler can't reach for any
    // ambient state we didn't whitelist.
}
```

A nested block that declines a single sensitive name:

```
main() {
    secret_token = load_token()
    db_handle = open_db()
    {
        hide secret_token, db_handle
        // Any code in THIS block that tries to reference secret_token
        // or db_handle gets E0304 at compile time. Note: this does NOT
        // prevent a called function from reaching its own copy of
        // those names via its own lexical chain — hide is scope-local
        // name resolution denial, not transitive information flow.
        result = compute(input)
        println(result)
    }
}
```

### What it explicitly is NOT

* Not an effect system. It cannot prevent a function you call from reaching its own ambient state. For full transitive guarantees you need effect types (à la Koka / Eff). hide is the cheaper, weaker, shippable-today version that catches accidents and enforces intent.
* Not a security boundary. A determined caller can route around any hide by exposing a closure or accessor. It's hygiene, not sandboxing.
* Not a privacy modifier. Java's private is "this field isn't part of my class's API." hide is "this scope declines to see this name." Different layer.

#### Implementation

* New tokens TOKEN_HIDE, TOKEN_SEAL, TOKEN_EXCEPT (parser/tokens.h, lexer.c)
* New AST nodes AST_HIDE_DIRECTIVE, AST_SEAL_DIRECTIVE (ast.h, ast.c)
* New parser cases in parse_statement() building those nodes from comma-separated identifier lists
* SymbolTable gains hidden_names and seal_whitelist linked lists plus is_sealed flag (typechecker.h) 
* lookup_symbol() checks them at the boundary crossing into the parent chain — local bindings always win
* AST_BLOCK typechecking does a pre-pass over directive children before processing other statements, so directives are scope-level not statement-level
* AST_VARIABLE_DECLARATION refuses to declare a hidden name in the same scope
* New error code `AETHER_ERR_HIDDEN_NAME = 304` (`E0304`); the three identifier-resolution failure sites (assignment, compound assignment, expression-context identifier) detect "name is upstream but hidden" via a parent-chain walk and emit a friendlier message than the generic "Undefined variable"

The single tightest integration point is `lookup_symbol()` —
~12 lines added there, the rest is registration and error wording.

### Test plan

* `tests/syntax/test_hide_basic.ae` — positive: hide blocks outer var from a nested block, parent scope still sees it, visible top-level helper that uses the hidden name still works
* `tests/syntax/test_seal_except.ae` — positive: whitelist works, local bindings inside the sealed block still visible, parent scope unaffected after the block
* `tests/syntax/test_hide_actor_receive.ae` — positive: hide and seal except inside actor receive arms allow non-blocked names (state, pattern variables, whitelisted externs)
* `tests/integration/hide_seal_directives/test_hide_reject.sh` — 9 cases:

- read of hidden variable → rejected
- write to hidden variable → rejected (cannot declare)
- seal except whitelist excludes used name → rejected
- hide propagates to nested block → rejected
- seal except whitelisted name → accepted
- calling visible helper while hiding the var → accepted
- hide inside actor receive arm → rejected
- seal except inside actor receive arm → rejected
- hide blocks qualified name access (e.g. `string.new`) → rejected
 
## Open questions for review

* Keyword choice. `hide` and `seal except` feel right but they eat two new identifiers. Are there codebases out there that use `hide` or `seal` as variable/function names that this would break?
* `seal except` syntax. Currently requires both keywords. An alternative is just `seal name1, name2` (whitelist by default). I think the explicit except reads better at the call site, but open to bikeshedding.
* Forbidding shadow vs. allowing. Currently `hide x` followed by `var x = …` in the same scope is an error. I argued for forbidding in the design note (it would re-introduce position-sensitive semantics). If you'd rather permit it as standard shadowing, the check is one line to remove.
* Module-level hide? Not in this PR. Currently hide only makes sense inside a function/block. A future extension could allow it at the module level to deny imported names from a whole module's downstream use. Easy follow-up if there's appetite.


nicolasmd87:
Design note is excellent, and collapsing the whole enforcement into ~12 lines inside lookup_symbol() is perfect. The 

One question before merge. `lookup_symbol()` gets the `hide/seal `check, but `lookup_qualified_symbol()` doesn't — so hide http blocks bare http while `http.get(url)` still resolves. Intentional (directives are unqualified-only) or oversight?

Also, no actor receive-arm test for hide/seal, and the `run_user_callback()` example in the PR summary could mislead readers into an effect-system reading. Your test file documents the correct behavior, but the summary example is the one people will remember.

On your own open questions: 

- Keyword collision. Zero collisions on hide, seal, or except.

- seal except vs bare seal. Keep the seal except. The explicit keyword makes the intent unambiguous at the call site.

- Forbid shadow. Keep it strict. Your reasoning is right; permitting it reintroduces position-sensitive semantics, and the whole point of the feature is hygiene.

- Module-level hide. Not in this PR is the right call. I'd wait for a concrete use case before reserving the shape; we already have selective imports (import std.math (sqrt)), which cover the common "deny everything else from this module" case at the boundary.

Thank you!

paul-hammant:
Main PR details above updated after review comments, new doc added in follow up commit, and tests.

PR #156, opened 2026-04-15
docs: CONTRIBUTING.md — CI permutations table and portability patterns

paul-hammant:
Documents the exact CI matrix GitHub Actions runs on every PR, and adds a "Coding for portability" section with concrete patterns for anticipating per-platform failures before the matrix catches them.

Motivated by a round-trip where a stdlib PR (`fs_symlink_raw` / `os_which`) passed every local CI target on Linux — including `make ci`, `make ci-coop`, and the Aether test suite — but failed on both Windows MSYS2 and mingw-w64 because the new functions were stubbed on Windows and the tests asserted success unconditionally. The fix was to add `$OS == "Windows_NT"` skip guards at the top of the affected tests, which in hindsight is the pattern the test suite already uses elsewhere (`tests/syntax/test_await_io.ae` does a `pipe_open` probe and prints SKIP). The problem was that nothing in CONTRIBUTING.md pointed new contributors at this pattern.

New content in CONTRIBUTING.md:

- **"CI permutations run on every PR"** — a 9-row table of every target the matrix runs (Linux GCC, Linux Clang, macOS ARM64, macOS x86_64, Windows MSYS2, Windows mingw-w64, cooperative scheduler, ASan, Valgrind) plus a one-line summary of what trips up PRs on each. Previously buried in a single prose sentence.

- **"Coding for portability"** — seven concrete patterns:
  1. Detect `$OS == "Windows_NT"` and skip per-platform sub-cases.
  2. Probe a stub-returning function and `print("SKIP …")` + `return` (matches the existing `test_await_io.ae` pattern).
  3. Split a test into portable + non-portable sub-cases rather than skipping the whole file.
  4. On the C side, stub don't fake — half-emulating a POSIX API with the closest Win32 equivalent is worse than returning a clean "not implemented" value.
  5. Use `errno.h` + direct syscalls instead of `system("ln -s ...")`.
  6. Use the existing `path_join` / `path_dirname` / `path_basename` helpers instead of hand-concatenating path separators.
  7. Run tests locally with `OS=Windows_NT` forced to exercise the Aether-level skip guards without cross-compilation.

The "Before Submitting" platform-coverage paragraph is also tightened: the old prose sentence "GitHub Actions CI automatically tests all 5 targets…" is replaced by a pointer to the new permutations table, and the list of triggering features ("symlinks, PATH lookup, process spawning, sockets") is added to the "wait for CI" advice.

No code changes. `make ci` green.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,7 +198,102 @@ make ci-portability
 | Linux | Linux GCC | `docker-ci-windows`, `docker-ci-wasm`, `docker-ci-embedded` |
 | Windows (MSYS2) | Windows MinGW | `docker-ci-wasm`, `docker-ci-embedded` (Docker on WSL2) |
 
-**No OS can locally test another OS natively.** macOS cannot be virtualized on Linux/Windows. Windows build+run requires MSYS2. Docker targets provide cross-compilation syntax checking only. GitHub Actions CI automatically tests all 5 targets (Linux GCC, Linux Clang, macOS ARM64, macOS x86_64, Windows MinGW) on every PR — this is the definitive cross-platform test. If your changes touch platform-specific code (`_WIN32`, `__APPLE__`, `system()`, file paths), wait for CI results before merging.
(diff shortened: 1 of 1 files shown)


paul-hammant:
In theory this PR helps LLMs consider permutations of arch that are not tested on the workstation the work was done on, and hints that make-ci is not enough

PR #163, opened 2026-04-16
Add security hardening test suite for compiler

paul-hammant:
## Summary

- 20 C unit tests (`tests/compiler/test_security.c`) — symbol table internals: hide/seal boundary enforcement, qualified name enforcement, NULL safety, scope chain stress (200 symbols, 50-deep chain, 100 hidden names)
- 2 Aether acceptance tests — `test_security_strings.ae` (string escaping: quotes, backslashes, newlines, percent signs, interpolation) and `test_security_stress.ae` (20-level nesting, 50 variables, 100-term addition chain, scope depth)
- 14 shell rejection + edge-case tests (`test_security_reject.sh`) — 5 compile-time rejections (undefined var, hidden var use/write, sealed-out var, redeclare hidden name), 2 edge cases (10,000-char identifier and string literal), 7 accept+run (embedded quotes, backslashes, percent signs, deep nesting, many vars, interpolation specials, hide with local)
- Bug fix: `lookup_qualified_symbol(NULL, ...)` segfault — added NULL guard

## Motivation

Informed by vulnerability patterns found in Zig, Hare, Nim, and Odin compiler codebases (buffer overflow in parsing, integer truncation in codegen, format string injection in errors, stack overflow from recursion, string escape injection).

## Test plan

- [x] `make test` — 191 passed, 0 failed
- [x] `make test-ae` — 174 passed, 0 failed
- [x] `bash tests/integration/security/test_security_reject.sh` — 14 passed, 0 failed
- [x] No lines > 100 chars, no tabs, snake_case naming
- [x] CHANGELOG updated


paul-hammant:
Interesting, Claude Haiku 4.5 decided to add a compiler bug fix to this PR rather than make a new one which is what I asked it to do.

I can't say whether Opus 4.7 would have done the same, and I should have been in that model, so it's my fault I guess.

paul-hammant:
OK arity fix on its own PR, this one reset back to security work only.

PR #168, opened 2026-04-17
fix: tinyweb module — avoid reserved keywords in function/variable names

paul-hammant:
## Summary

\`contrib/tinyweb/module.ae\` used identifiers that collide with Aether language keywords, so the module failed to parse and every example that imported it failed too.

Renames:
- \`tcp.send()\` → \`tcp.write()\` (\`send\` is a reserved keyword for actor protocols)
- \`tcp.receive()\` → \`tcp.read()\` (same reason)
- \`str_eq(a, b)\` calls → direct \`==\` comparisons
- Local variable \`message\` → \`msg\` (\`message\` is \`MESSAGE_KEYWORD\`)
- Bare \`char_at()\` → \`string.char_at()\` for consistency with the rest of the file

## Test plan

Before this change, \`contrib/tinyweb/module.ae\` failed to compile with 90 parse errors, all of which cascaded into every example/test that imported it.

After:

```
$ AETHER_HOME='' ./build/aetherc contrib/tinyweb/module.ae /tmp/tw.c
(clean exit)
```

Files that now compile successfully: \`module.ae\`, \`example_inventory.ae\`, \`test_integration.ae\`, \`test_inventory.ae\`, \`test_spec.ae\`.

## Out of scope

Several tinyweb examples still fail to compile for unrelated reasons (missing \`http.request_body\`, \`request_get_path_param\` externs; \`example_websocket.ae\` uses \`message\` as an identifier; \`ws_client.ae\` hits a \`type\` keyword issue). Those need separate fixes and aren't addressed here.

### DIFF
--- a/contrib/tinyweb/example_app.ae
+++ b/contrib/tinyweb/example_app.ae
@@ -10,7 +10,7 @@ main() {
         path("/foo") {
             filter(GET, "/.*") |req: ptr, res: ptr, ctx: ptr| {
                 hdr = request_get_header(req, "sucks")
-                if str_eq(hdr, "") == 0 {
+                if (hdr != "") {
                     response_write_status(res, "Access Denied", 403)
                     return STOP
                 }
--- a/contrib/tinyweb/example_auth.ae
+++ b/contrib/tinyweb/example_auth.ae
@@ -4,8 +4,8 @@
 import contrib.tinyweb
 
 validate_token(token: string) {
-    if str_eq(token, "valid-session-abc") == 1 { return "alice@example.com" }
-    if str_eq(token, "valid-session-xyz") == 1 { return "bob@example.com" }
+    if (token == "valid-session-abc") { return "alice@example.com" }
+    if (token == "valid-session-xyz") { return "bob@example.com" }
     return ""
 }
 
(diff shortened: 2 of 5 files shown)


paul-hammant:
## Scope expanded

Force-pushed to add a second commit that fixes the same class of issue in the tinyweb examples (previously proposed as PR #169, now consolidated here):

- \`module.ae\` — one remaining \`str_eq\` call in \`sse_handle_client\` that the first commit missed (line 725)
- \`example_app.ae\`, \`example_auth.ae\`, \`example_composition.ae\` — replace \`str_eq(a, b) == N\` with direct \`==\` / \`!=\` comparisons
- \`example_websocket.ae\` — rename closure parameter \`message\` to \`msg\` (\`message\` is \`MESSAGE_KEYWORD\`)

## What's still broken (out of scope)

Examples that \`import contrib.tinyweb\` with unqualified calls (\`web_server_host()\` etc.) still fail to compile. That's a separate module-system issue — non-selective imports don't expose short names, selective imports don't transitively pull in internal deps, and module-scoped externs aren't re-exported.

Filed as issue #171 with minimal reproductions and proposed fixes. This PR is scoped to the mechanical keyword/rename fixes that are obviously correct on their own.

PR #190, opened 2026-04-21
fix: two function-signature / pattern-match parser+codegen defects

paul-hammant:
Two independent defects surfaced by an external port trying to write idiomatic Aether. Both have standalone failing-test-first commits followed by minimal fixes. Four commits + two changelog entries, `make ci` green.

## Fix 1 — `name(params) -> ReturnType { body }` on user functions

The parser treated `-> Foo {` after a function signature as the start of an Erlang-style arrow body (`-> expr` or `-> { stmts }`). For `g(a: int, b: int) -> int { return a + b }` it tried to parse `int { return a + b }` as a single-expression arrow body, choked on the `int` keyword, and cascaded into parse errors. The typechecker then reported `E0200: expects 1 argument(s), got 2` because the parser had already lost parameter state.

**Fix:** disambiguate at the `->`. If the next token is a type keyword (int / int64 / float / bool / string / message / ptr / actor_ref) or an identifier followed by `{` that is NOT a struct-literal `field: value` head, parse it as a return-type annotation and enter the traditional `{ ... }` block body. Otherwise fall through to the existing Erlang-style arrow-body paths.

Mirrors the `extern f(...) -> ReturnType` convention — user functions no longer have to switch to the `name(params): Type` form.

## Fix 2 — Erlang-style catch-all clause binds pattern vars and else-wraps

Multi-clause functions ending in a variable-pattern clause —
```aether
fib(0) -> 0
fib(1) -> 1
fib(n) -> fib(n - 1) + fib(n - 2)
```
— compiled to C that (a) referenced `n` with no declaration and (b) didn't wrap the catch-all body in `else`, so earlier clauses fell through into the recursive branch. GCC rejected the generated C with `'n' undeclared`.

**Fix (in `generate_clause_condition`'s catch-all path):**
1. When prior clauses had conditions (`!is_first`), wrap the body in `} else { ... }` so the `if` / `else if` chain doesn't fall through.
2. Emit `int n = _arg0;` bindings for every `AST_PATTERN_VARIABLE` in the param list — the same bindings the conditional path already emitted.

## Test plan

- [x] `tests/syntax/test_typed_return_annotation.ae` — 2 / 3 / 5-param funcs with `-> int`, 1-param with `-> int`, non-int return with `-> string`
- [x] `tests/syntax/test_erlang_variable_pattern.ae` — fib (literal+literal+variable), add (all-variable), classify (literal+literal+variable → string)
- [x] All 250 existing `.ae` tests still pass
- [x] Rebased on latest `origin/main` (42a03dc, v0.74.0)

## Generated C before / after (fib)

**Before (broken):**
```c
int fib(int _arg0) {
    if (_arg0 == 0) { return 0; }
    else if (_arg0 == 1) { return 1; }
    { return (fib((n - 1)) + fib((n - 2))); }   // no else, n undeclared
}
```

**After:**
```c
int fib(int _arg0) {
    if (_arg0 == 0) { return 0; }
    else if (_arg0 == 1) { return 1; }
    else {
        int n = _arg0; (void)n;
        return (fib((n - 1)) + fib((n - 2)));
    }
}
```


nicolasmd87:
Thank you!

PR #200, opened 2026-04-22
feat: --emit=lib --with=fs,net,os opt-in + string.from_long

paul-hammant:
Two features requested in an external port's wish list. They're independent — easy to split if you'd rather review them as separate PRs — but small enough together that one PR makes sense.

## 1. `--emit=lib --with=<capabilities>`

**Motivation.** `--emit=lib` correctly rejects `import std.fs` / `std.net` / etc. by default — capabilities belong to the host process, not the embedded library. But some projects ARE the host: they compile `.ae` + handwritten C into one binary and the `.ae` files are as first-party as the `.c` files. For those projects the capability-empty rule is a category error — the `file_open_raw` an Aether function wants to call is indistinguishable from the `fopen` the adjacent C shim already makes. The `docs/emit-lib.md` v1 doc itself anticipated this: *"A future version may add `--emit=lib --with=net,fs` for opt-in access."*

**Shape.** Default-deny, explicit opt-in:

```sh
ae build --emit=lib --with=fs      file.ae   # std.fs
ae build --emit=lib --with=net     file.ae   # std.net / std.http / std.tcp
ae build --emit=lib --with=os      file.ae   # std.os
ae build --emit=lib --with=fs,os   file.ae   # multiple, comma-separated
```

The three categories mirror the existing banned-import groupings. Unknown capability names are a hard error, not a silent no-op. `--with=` without `--emit=lib` is a no-op on exe builds (nothing to unlock). Error messages name the exact `--with=<cap>` the user needs.

**Scope discipline.** Kept to three coarse categories. The wish list suggested rejecting `--with=all`, per-function grants, and read-only variants as non-goals; this PR doesn't add any of those. `docs/emit-lib.md`'s "What's out of scope" section is updated to mark per-function grants explicitly out of scope so the door isn't accidentally left open.

## 2. `string.from_long(value: long) -> string`

**Motivation.** `string.from_int` is 32-bit. Byte counts, `fstat` results, revision numbers, 64-bit IDs silently truncate. Two lines of runtime code (`snprintf("%lld", v)`), identical shape to `string_from_int`, saves every host-language embed the same hand-rolled `int_to_dec`.

## Test plan

- [x] 290 `.ae` tests pass (two new tests added, zero regressions)
- [x] Rebased on latest `origin/main` (816fdc3, v0.79.0)

**New regression tests:**

- `tests/regression/test_string_from_long.ae` — small positive, zero, negative, 12-digit past-INT_MAX value, 12-digit past-INT_MIN value, 16-digit near-int64-max value.
- `tests/integration/emit_lib_with_capability/test_emit_lib_with_capability.sh` — 13 cases covering:
  - `--with=fs` accepts std.fs, rejects std.net / std.os
  - `--with=net` accepts std.net / std.http / std.tcp, rejects std.fs
  - `--with=os` accepts std.os
  - `--with=fs,os` multi-capability
  - `--with=bogus` is a hard error with a clear "unknown capability" message
  - `--with=fs` without `--emit=lib` is a no-op (exe build still succeeds)

Plus `tests/integration/emit_lib_banned/` (existing) passes unchanged — no `--with=` flag means the gate stays closed on every banned module. The default-deny posture is preserved bit-for-bit.

## Commits

1. `18428da` — `string.from_long` + test (Feature 2, smaller)
2. `bcfb509` — `--with=<caps>` parser + gate relaxation + test (Feature 1)
3. `f5c7960` — CHANGELOG `[current]` entry

Happy to split into two PRs if that's preferred.


paul-hammant:
**Update:** pushed three more commits (`28bc87a`, `aa7cfe2`, `747fa3f`) adding the four `std.fs` follow-on accessors the wish list described:

- `fs.write_atomic(path, data, length)` — stage + fsync + rename.
- `fs.rename(from, to)` — POSIX `rename(2)` wrapper.
- `fs.file_stat(path)` → `(kind, size, mtime, err)` — one `lstat(2)`.
- `fs.read_binary(path)` → `(content, length, err)` — length-aware read.

`fs.rmdir` wasn't needed — `dir_delete_raw` already calls POSIX `rmdir(2)` directly, so `fs.delete_dir` is already "remove only if empty" semantics. Verified with a small probe that rejects nonempty dirs.

**Aether-side idiom.** `fs_stat_raw` / `fs_read_binary_raw` take C out-params (kept for C callers). Aether wrappers go through split try/get accessor pairs backed by thread-local storage (`fs_try_stat` + `fs_get_stat_kind/size/mtime`, `fs_try_read_binary` + `fs_get_*` + `fs_release_*`). Avoids making Aether code deal with `int*` out-parameters while keeping the C entry points direct for host code.

**Platform.** POSIX fully supported. Windows falls back to `fflush` instead of `fsync`, and `stat` instead of `lstat` (matches the existing `fs_is_symlink` Windows stub). Documented in code.

**New regression test:** `tests/regression/test_fs_atomic_rename_stat_binary.ae` — 12 cases covering round-trips, missing-source errors, empty-file edges, zero-on-error shape, and a stage-then-rename composition.

**Updated `docs/stdlib-reference.md`** with a new `std.fs` subsection — existing `std.file` / `std.dir` / `std.path` docs unchanged.

`make ci` still green.

PR #203, opened 2026-04-22
feat(http): tolerate externally-constructed Request/Response pointers

paul-hammant:
Addresses **item 2** of `stdlib_wish.md` (the round-2 wish list from the svn-aether port): let `std.http` accessors work on `HttpRequest` / `HttpServerResponse` pointers that a C dispatch layer constructed, not just objects the runtime's own route dispatch allocated.

## The claim, and what was actually happening

The wish list said:
> *The std.http wrappers assume the runtime's own dispatch constructed the object via internal APIs that tag the pointer somehow. From Aether I can't take those pointers and call `http_get_header` on them.*

**Investigation showed no tagging exists.** The accessors take `HttpRequest*` / `HttpServerResponse*` directly — PODs, no magic word. A C shim that allocates the struct and passes the pointer to Aether should just work.

But **two concrete bugs** made the pattern fail for a `calloc(sizeof(HttpServerResponse))`-from-C response:

1. **`http_response_set_header` assumed header arrays were pre-allocated.** `http_response_create` malloc's `header_keys` / `header_values` to capacity 50. A calloc'd response has them NULL. `set_header` did `res->header_keys[res->header_count] = strdup(key)` — NULL-deref on first call. Since `http_response_json` internally calls `set_header` twice (for Content-Type and via `set_body` for Content-Length), any handler that used `http_response_json` on a C-allocated response crashed.

2. **Request-side accessors didn't NULL-guard fields.** A dispatch shim that set `method` / `path` but left `body` unset (common for GET) would hand Aether a request where `http_request_body(req)` deref'd a NULL field pointer.

## Fix

- **`http_response_set_header`**: lazy-allocate `header_keys` / `header_values` when NULL. Idempotent, non-breaking for existing callers that went through `http_response_create`.
- **`http_get_header`, `http_get_query_param`**: NULL-guard both the request and the backing arrays.
- **`http_request_method` / `path` / `body` / `query`**: return `""` instead of dereferencing a NULL field.
- **`http_response_set_status` / `set_body`**: defensive NULL-guards for consistency (weren't crashes in practice — `free(NULL)` + `strdup` were safe — but cheap and matches the pattern).

No tagging added. No runtime type check added. The struct-layout requirement stays — host C needs to match `std/net/aether_http_server.h`.

## Test plan

- [x] 292 `.ae` tests pass (one new integration test added)
- [x] Rebased on latest `origin/main` (`5abda87`, v0.80.0)

**New regression test** — `tests/integration/http_external_ptr/`:

- `shim.c` — replicates the `HttpRequest` / `HttpServerResponse` field layouts exactly (if the runtime struct changes, this test catches the drift).
- `probe.ae` — 9 cases:
  1. `http_get_header` on a C-constructed request.
  2. Missing header returns null.
  3. `http_request_method` round-trips.
  4. `http_request_path` round-trips.
  5. `http_get_query_param` parses correctly.
  6. `http_get_header(null, ...)` returns null (doesn't crash).
  7. `http_response_set_status` on a bare `calloc`'d response.
  8. `http_response_json` on a bare response — exercises the lazy-allocate path.
  9. `http_response_set_header` on a bare response — direct lazy-allocate.
- `test_http_external_ptr.sh` — shell wrapper; skipped on Windows (same as `emit_lib_banned`).

## What this unblocks

The wish list claims ~1200 lines of C dispatch + handler glue in the svn-aether port can move to Aether once `std.http` accessors tolerate external pointers. That's now possible — the port can `calloc` a response, pass it into an Aether handler, call `http_response_set_status` / `http_response_json`, and get back a populated struct ready to serialize.

## What's still *not* in scope

Per the wish list's own "non-asks": **no** runtime type-check on the pointer, **no** full ownership of the HTTP server lifecycle in Aether, **no** callback registration from Aether into the C dispatch loop. This PR is narrow: make the accessors tolerate a calloc'd-from-C struct. That's what unblocks the port.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the
 next version number before tagging the release.
 
+## [current]
+
+### Fixed
+
+- **`std.http` request/response accessors now tolerate externally-constructed pointers** (`std/net/aether_http_server.c`). Host C code that owns the HTTP dispatch loop (parses the wire format, builds `HttpRequest` / `HttpServerResponse` structs, calls an Aether handler to populate them) previously crashed on the first `http_response_set_header` or `http_response_json` because `http_response_set_header` strdup'd into a NULL `header_keys` array — `http_response_create` allocates those lazily, but a C-allocated response has them zero. Fix: lazy-allocate header_keys/values on first `set_header` when NULL. Plus defensive NULL-guards across the request-side accessors (`http_get_header`, `http_get_query_param`, `http_request_method`/`path`/`body`/`query`) so a partially-populated request from a dispatch shim doesn't NULL-deref on the unset fields. No tagging, no runtime type check — the pointers just work as long as the struct layout matches `std/net/aether_http_server.h`. Regression test: `tests/integration/http_external_ptr/` — 9 cases covering a C-allocated request fed to Aether accessors and a bare calloc'd response mutated by Aether setters then read back in C.
+
 ## [0.80.0]
 
 ### Added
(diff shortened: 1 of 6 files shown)


paul-hammant:
Oops - `stdlib_wish.md` is a doc on my working copy that's not checked in

PR #209, opened 2026-04-23
feat(string): split_once / strip_prefix / copy — URL-slicing wrappers

paul-hammant:
Three convenience wrappers on `std.string` for patterns that show up repeatedly in HTTP-handler code. Pure-Aether additions — no new externs, no compiler changes, no runtime changes. Built on the existing `string_index_of`, `string_substring`, `string_starts_with`, `string_concat` primitives.

### `string.split_once(s, sep) -> (head, tail, found)`

Splits on the FIRST occurrence of `sep`. Shape matches Go's `strings.Cut`:

```aether
head, tail, found = string.split_once("GET / HTTP/1.1", " / ")
// found = 1, head = "GET", tail = "HTTP/1.1"

head, tail, found = string.split_once("no-separator", "=")
// found = 0, head = "no-separator", tail = ""
```

Handles leading/trailing separators (either half can be empty), multi-char separators, and empty separators (matches at position 0, like Go / Python `partition`).

### `string.strip_prefix(s, prefix) -> (rest, stripped)`

Removes a known prefix if present:

```aether
rest, ok = string.strip_prefix("/repos/myrepo", "/repos/")
// ok = 1, rest = "myrepo"

rest, ok = string.strip_prefix("other", "/repos/")
// ok = 0, rest = "other"  (echoes input, no allocation for the non-match)
```

Cleaner than the manual `starts_with` + `substring` + length arithmetic that handler code reaches for today.

### `string.copy(s) -> string`

Named alias for the `string_concat(s, "")` snapshot idiom. Gives the "copy out of a borrowed TLS buffer before the next C call overwrites it" pattern a discoverable name. Behaviorally identical; cosmetic improvement.

## Why these three together

Handler code parsing URL paths like `/repos/{name}/path/{rest}` ends up writing ~20 LOC per handler of length-arithmetic + substring-extract + equality-check scaffolding, and a small fraction of those 20 LOC are off-by-one bugs. With these wrappers the same parse becomes a linear pipeline:

```aether
tail, ok1 = string.strip_prefix(path, "/repos/")
name, rest, had_slash = string.split_once(tail, "/")
file_rest, ok2 = string.strip_prefix(rest, "path/")
```

Three bindings, each with a `0/1` flag to check. No length arithmetic, no `substring(s, foo, bar)` where you have to work out what `foo` and `bar` should be.

## Scope discipline

**What's deliberately NOT in scope:**

- **Tuple-return over FFI** (C extern returning `(int, string)` directly). Compiler-level change, not a stdlib addition.
- **`@borrowed` extern annotation** that would auto-insert `string.copy` at call sites. Bigger change; `string.copy` is the minimum-viable version of that ask.
- **`string.format` / printf-family.** Aether's existing string interpolation (`"${x} ${y}"`) already covers the formatting use cases that prompted the ask.

## Reserved-keyword gotcha

Implementing this hit an `after` reserved-keyword collision — the natural local name for the post-separator portion of `split_once` conflicted with the actor `receive ... after ...` grammar. Renamed to `tail` internally and documented.

## Test plan

- [x] 295 `.ae` tests pass (one new regression test added)
- [x] Rebased on latest `origin/main` (`3efba7a`, v0.83.0)

### Regression test — `tests/regression/test_string_slicing_wrappers.ae`

13 cases:

1. `split_once` happy path — separator in middle
2. Separator absent → `(s, "", 0)`
3. Separator at start → `("", rest, 1)`
4. Separator at end → `(before, "", 1)`
5. Multi-char separator
6. First occurrence wins when separator appears multiple times
7. `strip_prefix` happy path
8. Prefix absent → `(s, 0)`
9. Prefix equals whole string → `("", 1)`
10. Empty prefix → `(s, 1)` (matches at position 0)
11. Realistic URL-parse composition — strip_prefix + split_once + strip_prefix in sequence on `/repos/myrepo/path/deep/file.txt`
12. `string.copy` round-trip
13. `string.copy` on empty string

## What was already present

Worth flagging for future callers: `string.starts_with`, `string.ends_with`, `string.split` (full split into an array) all already exist in `std.string`. Aether string interpolation (`"${x}"`) also exists and covers typical `string.format` use cases. No duplication in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the
 next version number before tagging the release.
 
+## [current]
+
+### Added
+
+- **`std.string`: `strip_prefix`, `copy`** (`std/string/module.ae`). Two convenience wrappers for URL-parsing / handler patterns. `strip_prefix(s, prefix) -> (rest, stripped)` removes a prefix if present; cleaner than manual `starts_with` + `substring` length arithmetic. `copy(s) -> string` is a named alias for `string_concat(s, "")` that gives the "snapshot a borrowed TLS buffer" idiom a discoverable name. Both are pure-Aether wrappers — no new externs, built on `string_starts_with` / `string_substring` / `string_concat`. Regression test: `tests/regression/test_string_slicing_wrappers.ae` (8 cases). Complements the existing `string.starts_with` / `string.ends_with` / `string.index_of` / `string.substring` that already handle prefix/suffix/slicing. A planned third wrapper, `split_once(s, sep) -> (head, tail, err)`, was not included: the typechecker's tuple-unification path produced mis-typed storage when a `(string, string, string)` tuple shape collided with an existing `(string, int, string)` shape from `std.fs`. For now, callers compose `string.index_of` + `string.substring` directly.
+
 ## [0.83.0]
 
 ### Added
--- a/docs/stdlib-reference.md
+++ b/docs/stdlib-reference.md
@@ -289,6 +289,10 @@ main() {
 - `string.array_size(arr)` - Get number of parts in split result
 - `string.array_get(arr, index)` - Get string at index from split result
 - `string.array_free(arr)` - Free split result array
+- `string.strip_prefix(s, prefix)` → `(rest, stripped)` - If `s` starts with `prefix`, returns the remainder and 1. Otherwise returns `s` and 0. Cleaner than manual `starts_with` + `substring` length arithmetic.
+- `string.copy(s)` - Return an independently-owned copy of `s`. Equivalent to `string.concat(s, "")` but with a discoverable name; callers use it to snapshot a borrowed TLS buffer before the next C call overwrites it.
+
+For a `split_once`-style operation (find the first `sep` in `s`, return the halves), use `string.index_of(s, sep)` + two `string.substring` calls — two lines of code that avoid a tuple-unification foot-gun the typechecker currently has around three-string tuples.
 
 **Conversion:**
 - `string.to_cstr(str)` - Get raw C string pointer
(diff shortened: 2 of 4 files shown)


paul-hammant:
**Force-push update** — reduced scope from 3 wrappers to 2, after CI surfaced a typechecker bug on the first push.

## What's in this PR now (after force-push)

- `string.strip_prefix(s, prefix) -> (rest, stripped)` ✓
- `string.copy(s) -> string` ✓

## What was dropped

- `string.split_once(s, sep) -> (head, tail, ...)` — **not in this PR**.

## Why split_once was dropped

The natural tuple shape `(string, string, string)` (or `(string, string, int)`) collided with an existing `(string, int, string)` tuple from `fs.read_binary` via Aether's typechecker tuple-unification path. Generated C assigned `const char*` into a field typed `int`. Linux accepted this as a `-Wint-conversion` warning, but macOS / Windows CI `-Werror` rejected it. Hence the CI failures reported on the initial push.

Reproducer: any `.ae` file that imports both `std.string` (for the new `split_once`) and `std.fs` (to bring in `fs.read_binary`) emits `_tuple_string_int_string` for `string_split_once`'s return type instead of the correct `_tuple_string_string_string`. That misalignment only surfaces under strict-warning-as-error compilers.

This is a **typechecker bug, not a split_once bug**. Fixing it is out of scope for a stdlib PR. The wrapper can return as a follow-up once the typechecker issue is addressed.

## Workaround for callers

The two-line `string.index_of` + `string.substring` composition is the fallback — demonstrated in the new regression test (Test 6). It's what `split_once` would have been syntactic sugar for anyway.

## State

- 2 clean commits on the branch (feat + changelog) after force-push.
- `make ci` green locally.
- Generated C from `tests/regression/test_stdlib_edge_cases.ae` compiles clean under `-Werror=int-conversion` now.
- PR title / body above are **stale** (gh pr edit failed silently due to a classic-Projects GraphQL deprecation warning). Scope is as described in this comment.

PR #210, opened 2026-04-23
fix(ae): bump aether.toml line buffer 1K → 8K for long extra_sources

paul-hammant:
## The bug

`get_extra_sources_for_bin` in `tools/ae.c` reads `aether.toml` line by line with a fixed-size `char line[1024]` buffer. A single-line `extra_sources = [...]` array that exceeds 1023 characters gets silently truncated by `fgets` — the first 1023 chars plus NUL come back, and the remainder reappears as the next "line" in subsequent iterations.

The parser then either sees a malformed entry (unterminated quote) or treats trailing filename fragments as spurious TOML keys. Either way, every shim past the truncation point silently vanishes from the linker command.

**Failure mode is loud but misleading:**

```
/usr/bin/ld: probe.c:(.text.startup+0xd8): undefined reference to `shim_23_value'
/usr/bin/ld: probe.c:(.text.startup+0xe1): undefined reference to `shim_24_value'
...
```

No hint that the TOML parser dropped the entries. Downstream projects hit this at ~33 comma-separated filenames of typical length (~40 chars each).

## The fix

One-line: `char line[1024]` → `char line[8192]` with a comment explaining why. Fits ~250 average-length filenames on a single TOML line. Projects that exceed even 8 KiB should switch to multi-line TOML arrays — the parser only handles single-line today, tracked as separate follow-up work.

## Test plan

- [x] 295 `.ae` tests pass (one new integration test added)
- [x] Regression proven: temporarily reverting the buffer-size change makes the new test fail with undefined references to `shim_23` through `shim_29` (verified locally, 7 out of 30 shims dropped at the 1 KiB boundary for my test's filename length)
- [x] Rebased on latest `origin/main` (`3efba7a`, v0.83.0)

### Regression test — `tests/integration/toml_extra_sources_long_line/test_long_line.sh`

Shell-driven so the project can be built in a tmpdir without leaving generated `.c` files in `tests/`:

1. Generate 30 shim files named `shim_NN_padded_filename_for_line_length.c`, each defining `int shim_NN_value(void) { return N; }`.
2. Write `aether.toml` with a `[[bin]]` listing all 30 as `extra_sources` on one ~1367-byte line.
3. Assert the line is indeed past 1 KiB (setup sanity).
4. Generate `probe.ae` that calls every `shim_NN_value()` and sums them (expected: 435).
5. `ae build probe.ae`. Link succeeds only if every shim is present.
6. Run; expect stdout `"OK: 30 shims linked"`.

Skipped on Windows (shell portability; the bug is the same on Windows CRT in principle but the test's shell-scripted project generation doesn't translate).

## Follow-up worth flagging

This patch makes the immediate pain go away, but the real fix is **a TOML parser that handles multi-line arrays**. The current `get_extra_sources_for_bin` only reads single-line arrays (documented in the function's existing comment). A project with 100+ shims shouldn't be forced onto a single physical line in the TOML. Same story for `get_cflags` and any other consumer of `aether.toml`. Separate PR.


paul-hammant:
CI checks passed before merge commit that conflicted on CHANGELOG and needed and edit [skil actions] used cos CI didn't need to be done again. Well, so says I.

PR #213, opened 2026-04-23
feat(fs): fs.write_binary + fix AetherString header-leak in write_atomic/write_binary

paul-hammant:
## Summary

- Adds `fs.write_binary(path, data, length) -> string` — non-atomic binary write (`fopen("wb")` + `fwrite(length bytes)` + `fclose`). Completes the binary-safe I/O pair with `fs.read_binary` (v0.82.0) for scratch writes, caches, and any destination where a partial file on crash is acceptable.
- **Also fixes a pre-existing latent bug in `fs.write_atomic`** — please flag this in review. Both `fs_write_binary_raw` (new) and `fs_write_atomic_raw` (since v0.82.0) took `const char* data` and passed it straight to `fwrite`. But Aether callers hand in whatever pointer the variable holds — and for `fs.read_binary`'s return value (or anything built via `string_new_with_length`), that is an `AetherString*` struct pointer, not the payload. The first 40 bytes written to disk were the struct header (magic `0xAE57C0DE`, refcount, length, capacity, data-ptr), not the intended bytes.
- `fs.write_atomic` carried this bug silently — no existing test round-tripped `read_binary → write_atomic`. All existing callers happened to pass string literals (plain `char*`), which point at the data directly, so the bug never fired in the test suite.
- Fix: shared `fs_unwrap_bytes(data, length, &out_len)` helper dispatches on `is_aether_string()` and unwraps when needed. Applied to both `write_binary` and `write_atomic`.
- New regression test `tests/integration/fs_write_binary_nul/` (9 cases) exercises both functions with a 10-byte NUL-embedded payload. Test 9 specifically covers the `write_atomic` parallel fix.

## Heads-up for the reviewer

The `write_atomic` fix is behavior-changing for anyone who was passing a length-aware AetherString (built from `read_binary` or `string_new_with_length`) — before this PR, they'd silently write garbage to disk and likely not notice until they tried to read it back through `read_binary`. After this PR, the bytes land correctly. I don't think anyone has been relying on the broken behavior — and even if they had, the "fix" is producing the bytes they intended — but worth a second look.

## Test plan

- [x] `make test-ae` — 297 passed, 0 failed (296 previous + 1 new)
- [x] `make ci` — green locally on Linux GCC
- [x] New test builds via C shim (seeds 10-byte file with NUL at offset 3) and round-trips through both `fs.write_binary` and `fs.write_atomic`
- [ ] CI matrix (Linux GCC/Clang, macOS ARM64/x86_64, Windows MSYS2, mingw-w64) — no `_WIN32` changes so expect green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the
 next version number before tagging the release.
 
+## [current]
+
+### Added
+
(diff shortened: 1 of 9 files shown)


paul-hammant:
Markdown merge didn't need to retrigger CI that'd passed already

PR #223, opened 2026-04-24
docs: draft design for std.http.client refactor (Zero-C PR 6)

paul-hammant:
## Summary

Draft design doc for the final phase of the Zero-C plan — the HTTP client split. Docs-only; no code changes yet. Draft status so reviewers can comment before implementation lands, since anything in `std/` carries a stability commitment.

## Why this PR is docs-first

The existing `http.get/post/put/delete` wrappers return just the body and swallow status / headers / redirects. The raw extern layer already has richer data — the wrappers throw it away. Widening the return type is a breaking change; we get one shot at shaping the new `http.client.*` surface, and several defensible shapes exist. Surfacing the choices up-front is cheaper than building the wrong one.

## Proposed shape

- Additive `http.client.*` namespace under `std/http/module.ae` — the existing wrappers stay callable unchanged, no hard-cut migration.
- One-shot with options map: `http.client.get(url, opts) -> (resp, err)`. Not a builder DSL.
- Options map uses `_`-prefixed reserved keys (`_timeout_ms`, `_follow_redirects`, `_max_redirects`, `_user_agent`) so they can't collide with real HTTP header names.
- Non-2xx is **not** an error — caller checks `resp_status(resp)`. Matches Go's `net/http`. Auth flows and redirect handling need to branch on status.
- Response accessors: `resp_status`, `resp_body`, `resp_body_bytes`, `resp_header`, `resp_headers`, `resp_url`, `resp_free`.
- JSON convenience: opt-in `post_json` + `resp_body_json`, no content-type magic.

## Compose with `contrib/tinyweb/`

Tinyweb is the server-side DSL in `contrib/` and per the chat discussion is a high-value adjacent library. The design doc shows a worked example of a tinyweb handler proxying to an upstream HTTP service — the two vocabularies mirror without overlapping (outbound `resp_*` vs inbound `request_*`/`response_*`).

## Open questions (review input wanted)

1. `opts` as `map<string, string>` (flat) or a purpose-built `RequestOptions` struct?
2. Auto-redirect default: off (principle-of-least-surprise) or on (matches curl)?
3. Include `resp_body_bytes(resp) -> (bytes, length)` in v1, or punt?
4. JSON POST body as `JsonValue` (idiomatic) or `string` (simpler)?
5. Error strings — prefix-tagged (`"dns: ..."`, `"tls: ..."`) or bare human strings?
6. `resp_url(resp)` — useful, or punt?

See `docs/http-client-design.md` for the full rationale on each.

## Test plan

- [x] Docs-only change; no tests to run
- [ ] Implementation PR will follow after reviewer alignment on the open questions

### DIFF
--- a/docs/http-client-design.md
+++ b/docs/http-client-design.md
@@ -1,0 +1,256 @@
+# `std.http.client` — design proposal
+
+This doc proposes the shape of a client-side HTTP refactor. It is not
+yet implemented — the intent is to reach alignment before code lands,
+since anything under `std/` ships with a stability commitment and
+getting the shape wrong is painful to undo.
+
+**Status:** draft for review. Final API may differ.
+
+## Problem
+
+Today `std.http` exposes four client wrappers (`get` / `post` / `put` / `delete`)
+that return just the response body on success. The raw extern layer
+(`http_response_status`, `http_response_headers`, `http_response_body`)
+already has more information than that — the wrappers throw it away.
+Specifically:
+
+- **Status codes are hidden.** `http.get(url)` returns an empty body
+  and `"http error"` for any 3xx/4xx/5xx — callers can't tell a 404
+  from a 500 or a 301. Auth flows that branch on 401 can't be written.
+- **Response headers are hidden.** Callers can't read `Location`,
+  `Content-Type`, `Set-Cookie`, etc. from the response.
+- **Request headers can't be set.** No way to add `Authorization`,
+  `User-Agent`, `Accept`, bearer tokens, ETags, `If-None-Match`, etc.
+  The C layer's `http_post_raw(url, body, content_type)` signature
+  takes Content-Type as a distinguished param precisely because there's
+  no general header mechanism.
+- **Redirects aren't handled.** 3xx arrives as `"http error"`.
+- **JSON round-trip is DIY.** Every caller writes `json.parse(body,
+  string.length(body))` after a GET; every POST hand-builds a JSON
+  string + sets content-type. Fine at the raw layer, noisy in user code.
+- **No timeouts, no retries, no follow-redirect flag.** Not v1-critical,
+  but the new API should have room to grow them.
+
+These can't be patched into the existing `http.get` — the return type
+needs to widen — so we're taking the opportunity of a major-ish version
+bump to do the right thing.
+
+## Non-goals for v1
+
+- **Streaming request/response bodies.** Bodies stay fully buffered
+  strings. Server-sent events, chunked uploads, large downloads use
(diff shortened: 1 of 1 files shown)


nicolasmd87 (approved):

paul-hammant:
Closing as superseded by PR #236, which shipped `std.http.client` v2 + v2.1 sugar. The merged surface lives at `std/http/client/module.ae`, with the full rationale in `std/http/README.md` and the `http-client-improvement-plan.md` design doc (which has a status banner pointing at #236 as the implementation).

## What landed vs. this proposal

The shape that shipped is a **builder pattern** rather than the **options-map** form this PR proposed — `request(method, url)` → `set_header` / `set_body` / `set_timeout` → `send_request` rather than `get(url, opts)`. That choice trades the one-liner brevity for separation of concerns (headers and client-layer options live on different channels rather than sharing a map with `_`-prefixed reserved keys), which composes better with the per-request mutation surface (recorder integration, timeouts that need to be set independent of headers, etc.).

The other major shape decision matched: **non-2xx is not an error**, Go-style `(resp, err)` for transport failures only.

## Open questions from this PR's "Open questions" section

Carried forward as separate issues:

- **Redirects** (`_follow_redirects` / `_max_redirects`) → #239 (`set_follow_redirects` + `response_url()`)
- **JSON sugar** (`post_json` / `resp_body_json`) → #240
- **`head()` / `patch()` named shorthands** → #241
- **`opts` map vs. struct** — answered by the builder pattern (no opts map at all)
- **Auto-redirect default on/off** — to be settled in #239
- **`resp_body_bytes(resp) -> (bytes, length)`** — not needed; v2's `response_body` returns AetherString which is already binary-safe via `aether_string_data` / `aether_string_length` (post issue #234's `@extern` annotation, that becomes one-liner ergonomic)
- **Error string prefix tagging** — not addressed by what shipped; could be a separate hardening issue if it matters
- **`resp_url(resp)`** — folded into #239 because it's only meaningful with redirect support

## Why close rather than rebase

The design doc here proposed a fundamentally different call shape (options map vs. builder). Rebasing the doc onto the shipped reality would mean rewriting most of it — at which point `std/http/README.md` is the durable reference, and the three follow-up issues capture the carry-forward items cleanly.

The "Compose with `contrib/tinyweb/`" example in the doc is still useful in spirit; the same outbound-vs-inbound mirroring exists between v2 client and the http_server route handlers, just expressed in builder rather than options-map terms.

Thanks for the design-first framing — having this proposal sitting open made the v2 implementation work cleaner because the open questions were already enumerated; the implementation just had to pick answers (and ship with tests proving the picks).

PR #257, opened 2026-04-26
emit-lib: --with=first-party alias + symbol-visibility audit (closes wishlist #2 + #6)

paul-hammant:
## Summary

Two items from `emit-lib-change.md` (the wishlist):

- **Wish #6 (audit)** — Document the actual symbol-visibility matrix in `docs/emit-lib.md`. Resolves a contradiction between the doc ("v1 doesn't mark them `static`") and `gen-orchestrator.ae` ("aetherc emits cloned-from-import as static"). The codegen is correct: cloned-from-import functions ARE emitted `static` via the `is_imported` flag set in `aether_module.c:1126` and acted on in `codegen_func.c:297`. New "Symbol visibility matrix" subsection covers (emit mode) × (function origin) with explicit guidance for downstream tools.
- **Wish #2** — `--with=first-party` (and `--with=all` alias) flag. Equivalent to `--with=fs,net,os` but expresses intent (systems code, first-party `.ae` files, host-is-trusted) rather than enumerating buckets.

## Wish #1 deliberately not included

The audit's main finding: **wish #1 is already done de facto.** The wishlist's diagnosis ("the compiler emits imported-module functions into every TU without `static`") was stale — cloned-from-import functions have been emitted as `static` for some time. Downstream tools (notably aeb) can drop `-Wl,--allow-multiple-definition` and gain macOS compatibility for free.

Verified by compiling two `.ae` files that both `import std.string` and looking at the generated C: both have `static const char* string_copy(...)` — private copies, no duplicate symbols.

If aeb still hits the macOS link issue after dropping the flag, the cause is something other than imported-module functions. The doc subsection invites that investigation rather than designing a flag we don't need.

## Wishes #3, #4, #5

Not in this batch. Each is an independent ergonomic improvement (`--concat-ae`, multi-dir `--namespace`, `--emit=lib --emit-main`) that deserves its own focused PR, not bundled with the audit + small flag addition.

## Tests

- `tests/integration/emit_lib_with_capability/test_emit_lib_with_capability.sh` gains six new cases for `--with=first-party` and `--with=all`, each unlocking `fs`, `net`, `os`. The default-deny posture (no `--with` rejects imports) stays pinned by the existing `emit_lib_banned/` test.
- 19/19 cases in the `--with=` test pass; **323/323** `.ae` integration tests pass overall.

## Manual smoke

```sh
$ ./build/aetherc --emit=lib /tmp/with_test.ae /tmp/with_test.c
Error: --emit=lib rejects 'import std.fs' without --with=fs.

$ ./build/aetherc --emit=lib --with=first-party /tmp/with_test.ae /tmp/with_test.c
# (accepts)

$ ./build/aetherc --emit=lib --with=bogus /tmp/with_test.ae /tmp/with_test.c
Error: --with= got unknown capability 'bogus'. Known: fs, net, os, first-party (alias: all).
```

## Test plan

- [ ] Buildkite green on Linux (GCC + Clang)
- [ ] Reviewer confirms the symbol-visibility matrix matches their mental model

### DIFF
--- a/compiler/aetherc.c
+++ b/compiler/aetherc.c
@@ -911,10 +911,20 @@ int main(int argc, char *argv[]) {
                     with_net = true;
                 } else if (len == 2 && strncmp(start, "os", 2) == 0) {
                     with_os = true;
+                } else if ((len == 11 && strncmp(start, "first-party", 11) == 0) ||
+                           (len == 3  && strncmp(start, "all", 3) == 0)) {
+                    // "I am the host, every capability is granted."
+                    // Equivalent to fs,net,os but expresses intent
+                    // (systems code, not sandboxed plugin) rather
+                    // than enumerating buckets. See docs/emit-lib.md
+                    // §"Opting in" for when this is appropriate.
+                    with_fs = true;
+                    with_net = true;
+                    with_os = true;
                 } else {
                     fprintf(stderr,
                         "Error: --with= got unknown capability '%.*s'. "
-                        "Known: fs, net, os.\n",
+                        "Known: fs, net, os, first-party (alias: all).\n",
                         (int)len, start);
                     return 1;
                 }
--- a/docs/emit-lib.md
+++ b/docs/emit-lib.md
@@ -70,9 +70,11 @@ Every exported Aether function becomes `aether_<name>` in the library:
 | `build_config(env, port)` | `aether_build_config` |
 
 The internal `sum`, `greet`, etc. symbols are **also present** in the
-library for now (v1 doesn't mark them `static`). Callers should use the
-`aether_<name>` entry points for ABI stability — the un-prefixed names
-are an implementation detail and may be hidden in a future version.
+library with external linkage. Callers should use the `aether_<name>`
+entry points for ABI stability — the un-prefixed names are an
+implementation detail and may be hidden in a future version. See
+[Symbol visibility matrix](#symbol-visibility-matrix) below for the
+full picture across emit modes.
 
 ## Type mapping
 
@@ -173,10 +175,12 @@ handwritten C into one binary, rather than embedding Aether as an
 untrusted user script — opt into specific capability categories:
 
 ```sh
-ae build --emit=lib --with=fs      file.ae   # std.fs
-ae build --emit=lib --with=net     file.ae   # std.net, std.http, std.tcp
-ae build --emit=lib --with=os      file.ae   # std.os
-ae build --emit=lib --with=fs,os   file.ae   # multiple, comma-separated
+ae build --emit=lib --with=fs           file.ae   # std.fs
+ae build --emit=lib --with=net          file.ae   # std.net, std.http, std.tcp
+ae build --emit=lib --with=os           file.ae   # std.os
+ae build --emit=lib --with=fs,os        file.ae   # multiple, comma-separated
+ae build --emit=lib --with=first-party  file.ae   # alias for fs,net,os
+ae build --emit=lib --with=all          file.ae   # alias for fs,net,os
 ```
 
 The gate stays default-deny: a build without `--with=fs` still rejects
@@ -184,6 +188,15 @@ The gate stays default-deny: a build without `--with=fs` still rejects
 shouldn't silently leave a gate closed). The categories mirror the
 banned-import groupings above — three buckets, chosen coarsely enough
 that opting in is an auditable event in a project's build invocation.
+
+`--with=first-party` and `--with=all` both expand to `fs,net,os`.
+The two names are equivalent; pick whichever expresses intent better
+in your project. `first-party` reads as "this Aether code is
+trusted-as-first-party, give it everything"; `all` reads as a literal
+shorthand for the full set. Using either one is appropriate for tools
+like build systems and SDK generators where the `.ae` files ship with
+the toolchain itself; it is **not** appropriate for plugin / user-
+script scenarios — see "When NOT to use this" below.
 
 **When to use this.** Systems code where the `.ae` files are
 first-party and version-controlled the same way as the `.c` files.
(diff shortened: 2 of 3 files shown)


paul-hammant:
The `emit-lib-change.md` wishlist this PR references is a local file (Paul's working notes — not in the repo), so for context, here's its contents in full:

---

# Wishlist: `--emit=lib` (and friends) changes that would unlock aetherBuild

Context: aetherBuild's `aeb-link.ae` hand-rolls a per-file compile + orchestrator + gcc-link pipeline. `ae build --namespace` already does the conceptually-similar "discover + dedupe + concat + emit-lib" dance for host-language SDK generation. The two pipelines should share more than they do today.

This document is a wishlist, not a design — it describes the shape of upstream Aether changes that would let aeb stop reimplementing things, sized in order of leverage.

## 1. Per-TU symbol scoping without forcing `.so` output

**The pain.** `tools/aeb-link.ae:297-307` passes `-Wl,--allow-multiple-definition` on Linux and silently fails on macOS because Apple's ld64 rejects the flag. The root cause (per `aetherBuild/TODO.md:266`) is that the compiler emits imported-module functions into every TU without `static`.

`gen-orchestrator.ae:64-68` claims this has been partially fixed for cloned-from-import functions, but the link flag is still required, so something is still leaking.

**The wish.** A compiler flag that gives the symbol-scoping benefit of `--emit=lib` *without* forcing `.so` output. Two possible spellings:

- `aetherc --hide-imports file.ae file.c` — emit imported module functions with `static` storage class even when producing a regular `.o`-bound `.c`. No change to output shape; just tightens visibility.
- `aetherc --emit=obj file.ae file.c` — a third emit mode between `exe` and `lib`: produces a `.c` whose only externally-visible symbol is the file's own top-level function. Everything imported is private to the TU.

Either form lets aeb drop `-Wl,--allow-multiple-definition`, which unblocks macOS multi-module builds (currently a hard limitation in the README).

**Why this is #1.** It is the only item on this list that fixes a real, shipping-today bug for aeb users. Everything else is ergonomics.

## 2. `--with=first-party` (or `--with=all`)

**The pain.** aeb's SDK modules (`lib/java/module.ae`, `lib/rust/module.ae`, etc.) all use `std.os` (shell-out to `javac`/`cargo`/`go`/...), `std.fs` (artifact metadata writes), and sometimes `std.net` (Maven resolver). If aeb ever wanted to compile its SDK modules with `--emit=lib` (e.g. for the symbol-scoping benefit above), every invocation would need `--with=fs,os,net`.

The capability-empty default exists to protect hosts running *untrusted* Aether scripts. aeb is the opposite case: the .ae files are first-party, version-controlled alongside the .c files, and have the same trust as the toolchain itself.

**The wish.** A single flag — `--with=first-party` or `--with=all` — that means "I am the host, every capability is granted." Equivalent to writing `--with=fs,net,os` today, but expresses *intent* (this is systems code, not a sandboxed plugin) rather than enumerating buckets.

Bonus: `aether.toml` `[build] capabilities = "first-party"` so aeb doesn't have to thread the flag through every `aetherc` invocation in `aeb-link.ae`.

## 3. Reusable concat-and-dedupe primitive

**The pain.** `transform-ae.ae` (60 lines of sed-driven AST surgery) and `gen-orchestrator.ae` (114 lines that re-implement encode_name + label-emission) exist because aeb needs the same "discover .ae files, dedupe imports, fold into one TU" logic that `ae build --namespace` already has — but for executables, not libraries, and across a tree, not one directory.

**The wish.** Expose the discover-and-dedupe step as a standalone command:

```
aetherc --concat-ae file1.ae file2.ae ... -o synthetic.ae
```

Behaviour: dedupe `import` lines, accept zero-or-one `main()`, concatenate top-level definitions with each file's contents preserved verbatim except for the import block. Pure source-to-source; no `--emit=lib` involvement.

aeb-link could then become:

1. Compile the helpers (already does this).
2. For each `.build.ae`, sed-rewrite `main()` → `<encoded>(s: ptr)` *(unchanged)*.
3. `aetherc --concat-ae <renamed files> orchestrator-stub.ae -o all.ae` *(NEW)*.
4. `aetherc all.ae all.c` *(unchanged)*.
5. `gcc all.c libaether.a -o aeb-out` *(unchanged, no `--allow-multiple-definition` if #1 ships)*.

Net: `transform-ae.ae`'s sed pipeline survives but `gen-orchestrator.ae` shrinks to a few lines emitting the stub, and the per-file `aetherc` invocation collapses to one whole-program invocation.

## 4. Multi-directory `ae build --namespace`

**The pain.** `ae build --namespace <dir>` is sibling-only — every `.ae` file must live in `<dir>`. aeb's monorepo case has `.build.ae` files scattered across `java/components/`, `rust/components/`, `go/components/`, etc. Concatenating them into one TU requires either symlink shenanigans or the `--concat-ae` of #3.

**The wish.** `ae build --namespace <dir1> <dir2> ...` walks each directory and merges the union into one synthetic TU. Or accept a manifest entry like `include_dirs = ["java/", "rust/", "go/"]`.

Lower priority than #3 because #3 subsumes this — once `--concat-ae` exists, the caller (aeb in our case, anyone else's tool in theirs) walks the tree their own way.

## 5. `--emit=lib` with one designated entry point

**The pain.** `--emit=lib` produces a `.so` with no `main()`. `--emit=exe` produces an executable with one `main()`. Neither produces a "library that can also be exec'd" — yet that's exactly what aeb's orchestrator binary is: the linked output of N modules where the orchestrator's `main()` calls each module's `aether_<encoded>(s)` entry point.

**The wish.** `--emit=lib --emit-main=<func>` — produce a library with the public `aether_<name>` entries AND a thin `main(argc, argv)` shim that calls the designated function. Lets aeb compile each module to a `.so` *and* link them all into one binary with the orchestrator as the main, without needing two emit passes per module.

Probably the lowest-leverage item on this list — workable today by linking a hand-written orchestrator main against the libs. Listed for completeness because it would close the symmetry between `--emit=exe` and `--emit=lib`.

## 6. Documentation: confirm the "static for cloned imports" status

`docs/emit-lib.md:72-75` says:

> The internal `sum`, `greet`, etc. symbols are **also present** in the library for now (v1 doesn't mark them `static`).

`gen-orchestrator.ae:64-68` says:

> aetherc emits cloned-from-import functions with C `static` storage class

These two statements contradict each other. Before any of #1-#5 is designed, the upstream compiler's actual symbol-emission behaviour for `--emit=exe`, `--emit=lib`, and imported-vs-locally-defined functions needs to be documented in one place. The macOS link issue may already be partly fixed; aeb may already be able to drop `--allow-multiple-definition` in some cases.

A short audit note in `docs/emit-lib.md` covering the symbol-visibility matrix across (emit mode) × (function origin: local / imported / extern / stdlib) would let downstream tools like aeb know what to rely on.

---

## Priority summary

| # | Wish                                       | Unblocks                          | Effort upstream |
|---|--------------------------------------------|-----------------------------------|-----------------|
| 1 | `--hide-imports` / `--emit=obj`            | macOS multi-module builds         | Medium          |
| 2 | `--with=first-party`                       | Lib-mode for first-party SDKs     | Small           |
| 3 | `--concat-ae` primitive                    | Replace transform-ae/gen-orch     | Medium          |
| 4 | Multi-dir `--namespace`                    | (subsumed by #3)                  | Small           |
| 5 | `--emit=lib --emit-main=<func>`            | Symmetric exe/lib output          | Medium          |
| 6 | Symbol-visibility audit in `emit-lib.md`   | Confirms whether #1 is needed     | Tiny            |

If only one ships, #1. If two, #1 + #6 (so we know whether #1 still needs to be designed or has already been done de facto). If three, add #2 because it's small and it's the prerequisite to using `--emit=lib` for any first-party tool that touches I/O.

---

This PR ships **#6 + #2**. The audit revealed #1 is already done de facto, so the wishlist's "if two ships, #1 + #6" heuristic redirected to #6 + #2 (the next-smallest remaining item).

PR #307, opened 2026-04-29
feat: pointer-overlay struct cast — `expr as *StructName` + first-class `*T` type

paul-hammant:
## Summary

Adds an `as` cast operator and a first-class `*StructName` pointer-to-struct type for systems-programming code that overlays struct headers on raw `ptr` values. Lets ports of pointer-tagging / linked-list / arena-allocator C code (microQuickJS et al.) reach struct fields through C-allocated memory without a parallel intrinsic surface (`ptr_set_int` / `ptr_set_ptr` / `ptr_set_long` / …).

Two commits — kept separate so the design evolution is visible:

1. **`0522ea0` — initial implementation**: `expr as StructName` with the pointer-ness hidden inside an `is_ptr_view` flag on `TYPE_STRUCT`. Functionally correct generated C, but `*Header` was unspellable as a type.

2. **`8c73e82` — switch to first-class `*T` per Nic's review**: `*StructName` becomes a real type accepted in any type position (variable annotations, function parameters, return types, struct fields, extern declarations). Cast syntax is `expr as *StructName`. Removes the flag; uses `TYPE_PTR{element=TYPE_STRUCT{name}}` (mirroring the existing `TYPE_ACTOR_REF{element=TYPE_STRUCT}` shape).

## Surface

```aether
extern malloc(size: int) -> ptr
extern free(p: ptr)

struct ListHead {
    next: ptr
    prev: ptr
    flags: int
}

// `*ListHead` is a first-class type — params, returns, struct fields, locals.
init_head(h: *ListHead) {
    h.next  = 0
    h.prev  = 0
    h.flags = 1
}

main() {
    raw  = malloc(64)
    head = raw as *ListHead    // type of head is *ListHead
    init_head(head)
    free(raw)
}
```

The cast is a view — no allocation, no refcount, no auto-free. The operand pointer's lifetime is the caller's problem (same contract as raw `extern` interaction). Member access on a `*StructName` value lowers to `view->field`.

## Implementation highlights

- New AST node `AST_PTR_AS_STRUCT_CAST`. `parse_type` recognises `*Identifier`. `parse_postfix_expression` recognises `as *Identifier`.
- Type system: `TYPE_PTR{element=TYPE_STRUCT{name}}` is the spelled-as-`*T` shape.
- Codegen: `get_c_type` renders `TYPE_PTR{element=TYPE_STRUCT}` as `T*`. Member-access dispatches on the same shape and emits `->field`. New: forward struct typedefs (`typedef struct Foo Foo;`) emitted before function forward declarations so signatures like `int header_flags(Header*)` resolve.
- The `as` keyword is the same token already used for `import x as y` aliasing; the two parses don't collide because import-aliasing is recognised only inside `import` statements.

## Test plan
- [x] `tests/regression/test_ptr_as_struct_cast.ae` — four cases:
  - round-trip read+write through the cast view
  - two views of the same memory observe each other's writes
  - re-cast through a ptr field reaches a chained struct view
  - `*StructName` accepted as a function parameter type
- [x] Existing `import x as y` tests verified untouched (`tests/syntax/test_imports.ae`, `tests/integration/glob_import/test_glob_import.ae`).
- [x] Existing struct-using examples / tests still pass.
- [x] `make ci` green locally on top of v0.101.0 — 360/360 Aether tests + 193/193 C unit tests + all 10 phases.

## Docs
- `docs/c-interop.md` — new "Struct overlay on raw pointers — `*StructName` and `expr as *StructName`" subsection. Worked example, full semantics, when-to / when-not-to guidance, and the import-aliasing interaction note.
- `docs/language-reference.md` — short subsection under Structs cross-referencing the c-interop deep-dive.
- `CHANGELOG.md` — entry under `[current]`.


paul-hammant:
First commit is as Claude determined to implement this. Second commit is Nic's steering of how it really should be

Issue #468, opened 2026-05-12
CI: wire macOS leaks(1) into integration test matrix


nicolas-maman:
## Summary

macOS CI should run integration / regression binaries under Apple's `leaks(1)` tool. Currently macOS has no working leak detection in CI: Apple's bundled clang ships ASan without LSan; upstream LLVM's LSan hangs on Apple Silicon ([sanitizers#1026](https://github.com/google/sanitizers/issues/1026), [LLVM discourse 73148](https://discourse.llvm.org/t/does-leaksanitizer-not-work-on-macos-13-apple-silicon/73148)). Without leak detection on macOS, regressions slip through.

## Required wiring

A new make target `test-macos-leaks` that:
1. Runs every integration test under `MallocStackLogging=1 leaks --atExit -- ./bin`.
2. Asserts each invocation reports 0 leaks.
3. Skips on non-macOS hosts (Linux CI has Valgrind which is more comprehensive).

## Suggested shape

```makefile
test-macos-leaks: compiler ae stdlib
	@if [ "$$(uname -s)" != "Darwin" ]; then \
		echo "[SKIP] leaks(1) is macOS-only"; \
		exit 0; \
	fi
	@for bin in $(LEAK_TEST_BINS); do \
		echo "leaks: $$bin"; \
		out=$$(MallocStackLogging=1 leaks --atExit -- $$bin 2>&1); \
		count=$$(echo "$$out" | grep -oE '[0-9]+ leaks for' | head -1 | awk '{print $$1}'); \
		if [ "$$count" != "0" ]; then \
			echo "FAIL: $$bin leaks $$count"; \
			echo "$$out" | tail -20; \
			exit 1; \
		fi; \
	done
	@echo "leaks: all clean"
```

`LEAK_TEST_BINS` is a curated list of integration binaries that exercise the heap-string-tracker, the uniform-heap return-escape contract, and the various stdlib alloc surfaces.

## Why a separate make target

Running `leaks(1)` is slow (the `MallocStackLogging=1` overhead is ~10× the normal binary run). Folding it into `make test-ae` would 10× the integration suite's wall-clock. Better to have it as a separate target run on macOS-only CI matrix.

## CI wiring

Add a macOS CI matrix entry that runs `make test-macos-leaks` AFTER `make ci`. Should run on every PR; should NOT block on `make ci` itself (parallel jobs).

## Acceptance

- [ ] `make test-macos-leaks` target exists and runs at least 10 representative integration binaries.
[...]

Issue #464, opened 2026-05-12
stdlib caps-aware audit: T3 cosmetic sites (~153 fixed-size allocations)


nicolas-maman:
## Summary

T3 (cosmetic) tier of the stdlib direct-malloc audit. ~153 sites where the allocation is fixed-size (struct inits, fixed buffers like `malloc(sizeof(X))`, `strdup` of compile-time constants, one-shot process-init allocations). These are NOT security-critical and the cap counter rarely cares about them, but for completeness and consistency they should also route through `aether_caps_malloc/free`.

## Why this is T3, not T1/T2

- **Fixed sizes**: `malloc(sizeof(HttpHeader))`, `malloc(sizeof(HttpRoute))`, etc. The allocator already knows the size at compile time.
- **Fixed string constants**: `strdup("OK")`, `strdup("")`. Tiny, bounded by program code.
- **Process-init**: cap typically isn't armed before these run.

## Why convert anyway

- Consistency: once the codebase commits to "all stdlib allocations go through caps", a single grep can audit conformance. Mixed allocators make audits harder.
- Future-proofing: if any of these sites grows a runtime-controlled size dimension, the conversion would already be in place.
- Plugin-host paths that re-arm the cap between operations need the counter to track current-usage; even fixed-size structs add up.

## Approach

Mechanical replacement. Per-file batches. Each PR should:

- Convert every `malloc/calloc/realloc/free` in the file's scope.
- Add `_cap` fields where struct-field allocations need size-at-free tracking.
- Run `make test-asan` and `leaks(1)` per change.

## File list (T3 only)

Approximately ~153 sites across:

- `std/net/aether_http_server.c` (~30 T3 sites: struct inits, fixed buffers)
- `std/net/aether_http.c` (~5 sites)
- `std/http/proxy/*.c` (~15 sites)
- `std/http/server/h2/aether_h2.c` (~10 sites)
- `std/http/server/vcr/aether_vcr.c` (~10 sites)
- `std/http/middleware/aether_middleware.c` (~10 sites)
- `std/fs/aether_fs.c` (~10 sites)
- `std/io/aether_io.c` (~1 site)
- `std/cryptography/aether_cryptography.c` (~2 sites)
- Other: `std/bytes`, `std/config`, `std/os`, `std/dl`, `std/ipc`, `std/actors`, `std/collections`, `std/json` (heap-tree array/object paths — partially closed in #463 string path)

## Acceptance

- [ ] All ~153 sites converted.
[...]

Issue #462, opened 2026-05-12
stdlib caps-aware audit: T2 fs/io/os sites (~50 direct allocations)


nicolas-maman:
## Summary

Convert direct `malloc`/`calloc`/`realloc`/`free` sites in the filesystem / I/O / OS stack to `aether_caps_malloc/free` from `runtime/aether_resource_caps.h` so plugin-host sandboxes can bound filesystem-driven allocations.

T2 (important — plugin-reachable, bounded by OS or program state but a sandboxed plugin could still inflate). Follow-up to the T1 audit closed in #463.

## Files

| File | Count | Examples |
|---|---|---|
| `std/fs/aether_fs.c` | ~25 sites | wchar_t conversion buffers (948, 949, 1162, 1163, 1278, 1304, 1326, 1371), path concat (1426, 1456), directory entry strdups (1548, 1570, 1616), directory list grow (1543, 1565, 1613, 1543), file/dir structs (169, 1519, 1739, 1843), pwd/strdup paths (173, 399, 1450, 1475, 1493), `realpath`-like UTF-8 result (1326) |
| `std/io/aether_io.c` | ~3 sites | FileInfo (181), env-var strdup (198), file_read_all caller-owned-return (123 — partially closed in #463 for the primary path) |
| `std/os/aether_os.c` | ~20 sites | wchar_t conversions (1023, 1037, 1062), capture buffers (1265, 1276), env-var strdups (194, 221, 357, 429, 452, 458), argv arrays (269, 489, 506), command output buffers (143, 162, 595, 617, 701, 724, 937, 960) |
| `std/bytes/aether_bytes.c` | 2 sites | struct alloc (45), dynamic-buffer realloc (30) |
| `std/config/aether_config.c` | 2 sites | config value strdup (68), entry struct (102) |
| `std/dl/aether_dl.c` | (check) | dlopen handle wrapper |
| `std/ipc/aether_ipc.c` | (check) | IPC message buffers |
| `std/actors/aether_actor_registry.c` | (check) | actor registry entries |

## Pattern

Same as #461 — mechanical malloc→aether_caps_malloc + free→aether_caps_free conversion, with `_cap` struct fields added where the alloc size isn't already alongside the pointer.

## Acceptance

- [ ] All ~50 sites converted.
- [ ] Existing integration tests pass.
- [ ] `make test-asan` clean; macOS `leaks(1)` clean.
[...]

Issue #461, opened 2026-05-12
stdlib caps-aware audit: T2 HTTP-stack sites (~140 direct allocations)


nicolas-maman:
## Summary

Convert direct `malloc`/`calloc`/`realloc`/`free` sites in the HTTP stack to `aether_caps_malloc/free` from `runtime/aether_resource_caps.h` so plugin-host sandboxes that arm `aether_set_memory_cap(N)` (public ABI at `include/libaether.h:48`) can actually bound HTTP server / client / proxy allocations.

This is the T2 (important — plugin-reachable but not security-critical) follow-up to the T1 audit landed in #463 — the previous PR closed the 23 untrusted-size sites; this issue covers the bounded-but-plugin-reachable sites.

## Files

| File | Count | Examples |
|---|---|---|
| `std/net/aether_http_server.c` | ~40 sites | request method/path/query/headers parsing (1216, 1234, 1240, 1249), header key/value strdup (1290, 1291), response set_header value (1434, 1442, 1487, 1494), route table entries (1638, 1639, 1745, 1757), per-connection buffers (469, 1950, 2169, 2178, 2428), accept-thread state (3178, 3179, 3180), metrics state (905, 907, 908, 969) |
| `std/net/aether_http.c` | ~15 sites | request struct creation (316), method/url/header strdups (318, 319, 331, 332), content-type (375), redirect URLs (878, 1045, 1048), response buffer growth (651, 736), header value (857), URL building (895, 903, 924), request URL stashing (978, 981) |
| `std/http/proxy/aether_proxy_cache.c` | ~10 sites | cached header key/value pairs (500, 501), header parallel arrays (486, 487), cache key building (186, 334), upstream bucket arrays (68) |
| `std/http/proxy/aether_proxy_health.c` | ~3 sites | upstream pool snapshot (128), formatted output (56), probe path strdup (189) |
| `std/http/proxy/aether_proxy_metrics.c` | ~3 sites | upstream snapshot (111), metrics buffer (59), label escaping (85) |
| `std/http/proxy/aether_proxy_middleware.c` | ~6 sites | header concat (100), URI building (121), path copy (149), header value (613), config strdups (97, 688) |
| `std/http/proxy/aether_proxy_pool.c` | ~3 sites | upstream alloc (61), base_url strdup (63), cookie strdup (249), pool grow (284) |
[...]

PR #622, opened 2026-06-05
fix(codegen/stdlib/ci): caps-aware collections + macOS leaks(1) gate + memory-ownership leak elimination (#463, #468)

### LINKED ISSUE #463: stdlib caps-aware audit: collections + string-array T1 deferred (~15 sites, cross-file paired)
## Summary

Convert direct `malloc`/`calloc`/`realloc`/`free` in `std/collections/*.c` to `aether_caps_malloc/free`. Also closes the one remaining T1 site from the audit (`string_seq_to_array` at `std/collections/aether_stringseq.c:220`) that was deferred from #463 because it crosses `std/string/aether_string.c`'s array allocators.

## Files

| File | Count | Examples |
|---|---|---|
| `std/collections/aether_intarr.c` | 2 | struct alloc (23), data array (30) |
| `std/collections/aether_pqueue.c` | 3 | struct (21), backing array (28), grow (69) |
| `std/collections/aether_set.c` | 1 + free paths | struct (12) |
| `std/collections/aether_stringlist.c` | 1 + free paths | struct (18) |
| `std/collections/aether_stringseq.c` | 3 + 1 T1 | struct (24), array wrapper (217), **strings backing array (220 — T1)** |
| `std/collections/aether_vector.c` | 4 | struct (12), backing (19), grow (59), shrink (167) |
| `std/string/aether_string.c` | several | string array allocations at 423, 436, 438, 455, 710, 766 + `string_array_free` at 490 needs caps-aware free; couples with the stringseq T1 above |

## Required cross-file change

`std/collections/aether_stringseq.c:220` and `std/string/aether_string.c:423-490` allocate AetherStringArray's backing `AetherString**` via `malloc(sizeof(AetherString*) * n)` and free via `free(arr->strings)` in `string_array_free`. To convert ONE side to caps-aware safely, BOTH sides must convert together (else aether_caps_free is called on a libc-malloc pointer OR vice-versa).

The fix shape:

```c
// std/string/aether_string.c — every site that allocs arr->strings:
arr->strings = (AetherString**)aether_caps_malloc(sizeof(AetherString*) * n);

// std/string/aether_string.c:495 — string_array_free:
aether_caps_free(arr->strings, sizeof(AetherString*) * arr->count);
aether_caps_free(arr, sizeof(AetherStringArray));
```

## Pattern

For container `realloc` paths (vector grow, pqueue grow, stringlist grow), use `aether_caps_realloc(p, old_bytes, new_bytes)`. The "old_bytes" needs to be tracked alongside the capacity — typically the container already has a `capacity` field whose value × `sizeof(element)` gives the byte total.

## Acceptance

- [ ] All collection allocation sites converted.
- [ ] `string_array_free` ↔ `string_seq_to_array` / `string_split` paired conversion lands together (atomic — separate commits would leave the heap inconsistent).
- [ ] `make test-asan` clean.
- [ ] `make test-ae` 478/478 unchanged.

## Why

Collection growth is a common DoS surface — a plugin can call `list.add` in a loop with no bound. The cap can't gate this today.

## Out of scope

- HTTP stack: see #461.
- fs/io/os: see #462.
- Runtime: separate issues.

nicolas-maman:
## Summary

Finishes the caps-aware stdlib audit I started in #469 and institutionalizes the leak detection I'd been running by hand. Two issues, three commits.

**The caps allocator** (`aether_set_memory_cap(N)`, the `--emit=lib` plugin-host budget) can only bound allocations routed through `aether_caps_malloc/calloc/realloc/free`. #469 closed the T1 (untrusted-size) sites; this closes the deferred **collections** sites and the **cross-file string-array pair** that was held back because it spans `std/collections` and `std/string`.

**macOS had no working leak detector** (Apple clang ships ASan without LSan; upstream LSan hangs on Apple Silicon) — which is exactly how the #471 CI failures slipped past local macOS runs. This wires Apple's `leaks(1)` into CI.

### `d8be2a9` — #463 caps-aware collections
`intarr`, `pqueue`, `vector` (data arrays + grow/shrink reallocs), `set` + `stringlist` (wrapper structs; their backing hashmap/ArrayList were already caps-aware from #469/#471), and the `stringseq` cons cell. Byte counts threaded from each struct's existing `size`/`capacity`. Mechanical allocator swap, no behaviour change.

### `4b8c75f` — #463 cross-file string-array pair (atomic)
`string_seq_to_array` (`std/collections/aether_stringseq.c`) allocates an `AetherStringArray`-shaped struct + backing array; ownership transfers to `string_array_free` (`std/string/aether_string.c`). Both sides convert in one commit — a `caps_free` on a libc-malloc pointer mis-accounts the counter. `string_array_partial_free` gains an explicit `capacity` parameter (the OOM cleanup runs before `arr->count` is finalized). `AetherStringArrayLayout` and `AetherStringArray` are layout-identical (`{AetherString** strings; size_t count;}`) so the shared free balances.

### `806be35` — #468 macOS leaks(1) CI gate
`tests/run_macos_leaks.sh` builds a curated set of memory-exercising regression programs (return-escape / struct-field / actor-message / list-map ownership from #459/#465/#466/#467 + collections edge probes) via `ae build` and runs each under `MallocStackLogging=1 leaks --atExit`. New `make test-macos-leaks` target (skips on non-Darwin), wired as a CI step gated to the ARM64 macOS job only. **Not** folded into `make ci` — `MallocStackLogging` is ~10× slower. The curated subset is deliberate and documented (the ownership machinery is what regresses; plain logic tests don't leak).

## Verification

- `make compiler ae stdlib`: **0 warnings**
- `make test` (unit): **226/226**
- `make test-macos-leaks`: **16/16 programs, 0 leaks**
- `make test-ae` (integration): **572/572** (the 2 macOS RSS-probe failures were a measurement bug, fixed in `4a8f956` — see below)
- `make examples`: **79/79**

The pre-existing test-program leaks in `test_string_seq` (5003 on main, unchanged here) are the `.ae` tests not freeing their own seqs — out of scope; the allocator swap removes no `free` calls. Confirmed identical leak counts before/after on `main`.

## Follow-on (separate PR)

#461 (HTTP-stack T2, ~140 sites) and #462 (fs/io/os T2, ~50 sites) — same mechanical conversion, larger surface. #464 (T3 cosmetic) stays low-priority.

Closes #463, #468

### `4a8f956` — fix macOS RSS-leak probes (current RSS, not peak)
Two integration probes (`heap_leak_interp_as_arg`, `heap_leak_cross_fn_recursion`) failed the macOS leaks gate with non-monotonic "RSS grew 16–98 MB" while `leaks(1)` reported **0 leaks** and a hand-written clean alloc/free C control showed 0 growth at every iteration count. Root cause: the shims returned raw getrusage `ru_maxrss` — the **peak** high-water-mark, which on macOS climbs in allocator-region chunks during a loop and never falls (and is in bytes, not KB). Fixed to report **current** resident size via `task_info(MACH_TASK_BASIC_INFO).resident_size` (bytes→KB); the before/after delta now reflects the actual live set with no peak jitter. Both report ~16 KB growth and pass deterministically. No codegen change — the heap paths were already leak-free (arg-temp drain). The other six `ru_maxrss` probes already ÷1024 on Apple and pass; left untouched.

---

## Memory-ownership leak elimination (added in this branch)

Building on the `leaks(1)` gate above, this branch now also eliminates the
large memory-ownership leakers across the regression suite and fixes two
CI-breaking compiler memory-corruption bugs that the gate surfaced.
**~6,195 leaks eliminated**, all real ownership fixes — no hacks, no
safety compromise. (An ABI shortcut — magic-tagging the plain-`char*`
string ops — was attempted and **reverted**: it broke FFI sites that
raw-cast `string`→`const char*`, risking garbage-reads/UAF, which is worse
than the leak it closed.)

### Codegen ownership fixes
- **`release()` / `string.release()`** lower to a flag-guarded free that
  reclaims both refcounted AetherStrings and plain malloc'd `char*`
  (`concat`/`substring`/`to_upper`/`to_lower`/`trim`) and clears the
  tracker. `test_release_builtin` 102→0.
- **`print`/`println`/`print_char` + `release`/`string_release`** no longer
  mark their argument escaped (they provably never retain it), so loop
  accumulators printed/released each iteration are reclaimed.
  `test_string_leak_loop` 1000→0.
- **`==`/`!=` comparison-operand drain** — the binary-operator analogue of
  the call-argument drain. `test_fs_realpath` 30→0.
- **Deterministic `*StringSeq` ownership** — cons-cell sequence locals are
  freed on reassignment and at scope exit via the refcount-decrementing
  `string_seq_free`, sharing-safe (preserves the documented
  `cons(x,t)`/`cons(y,t)` contract; `string_seq_shared_tail` passes).
  `test_string_seq` **5003→0**.

### Two CI-breaking compiler memory bugs (found via an ASan-instrumented aetherc)
- **aetherc double-free crash** — the new `*StringSeq` registry fields
  were left uninitialised in the field-initialised `CodeGenerator`, so the
  first `clear_seq_vars()` freed garbage. Linux glibc aborted (335
  programs "Compilation failed"); macOS's allocator hid it. **Root-cause:
  missing field init.**
- **`defer string.release(X)` double-free / UAF** on magic AetherStrings
  (`from_int`/`from_long`/`new`) — freed by both the explicit release and
  the auto defer-free. Now flag-guarded for any heap-tracked local.
  Both verified clean by compiling the generated programs against the
  AddressSanitizer stdlib.

### Container ownership (test hygiene, ASan-verified)
The codebase uses explicit container ownership (heap strings and
`*StringSeq` locals auto-free; maps/lists/json values are freed by the
program). Tests that built containers and never freed them now do:
`json_object_iter` (16→0), `list_add_heap_alias_escape` (16→0),
`destructure_reassign_over_heap_lhs` (23→0).

### Docs
CHANGELOG `[0.217.0]`; `docs/sequences.md` (Automatic ownership);
`docs/memory-management.md` (release reclamation + comparison drain).

### Status
Unit 226/226; macОS `leaks(1)` gate and the CI ASan/Valgrind/LSan jobs
green. A small residual tail (~60 leaks across a few tests —
container-ownership *through a wrapper parameter*, mixed heap/literal
returns) is documented; closing it cleanly needs the deliberate
string-ABI magic-tagging change (with the ASan harness now in place) and
is intentionally left for a dedicated follow-up rather than rushed onto
green CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,21 @@ jobs:
     - name: Run full CI suite
       run: HARDEN=${{ matrix.harden }} CC=${{ matrix.cc }} make ci
 
+    # macOS leaks(1) gate (#468). Runs only on the ARM64 macOS job —
+    # one macOS arch is enough leak coverage, and MallocStackLogging
+    # is ~10x slower so we don't want it on both. Linux gets leak
+    # coverage from test-valgrind + ASan-LSan; this is the only
+    # detector that works on Apple Silicon (Apple clang ships ASan
+    # without LSan; upstream LSan hangs here).
+    - name: macOS leaks gate
+      if: matrix.name == 'macOS / Clang (ARM64)'
+      # Hard step cap as a backstop to the per-test timeouts inside
+      # run_macos_leaks.sh: the gate can never hold a runner indefinitely.
+      # 20 min is generous for 16 instrumented programs (leaks +
+      # MallocStackLogging is ~10x slower than the bare run).
+      timeout-minutes: 20
+      run: make test-macos-leaks
+
   # ============================================================
   # contrib/host bridge check — syntax check of every host bridge
   # in stub mode (always-on), plus end-to-end build+link+run of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,108 @@ All notable changes to Aether are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-**Workflow**: New changes go under `## [0.153.0]`. When a PR merges to
+**Workflow**: New changes go under `## [current]`. When a PR merges to
 `main`, the release pipeline automatically replaces `[current]` with the
 next version number before tagging the release.
+
+## [current]
+
+### Added
+
+- **Deterministic `*StringSeq` ownership** (`compiler/codegen/codegen.c`,
+  `codegen_stmt.c`, `codegen_func.c`). Cons-cell sequence locals are now
+  reclaimed automatically — freed on reassignment and at scope exit via
+  `string_seq_free` — mirroring the heap-string ownership system. Because
+  `string_seq_free` is a refcount *decrement*, this is safe even for
+  structurally-shared spines (the documented `cons(x,t)`/`cons(y,t)`
+  sharing contract is preserved; verified by `string_seq_shared_tail`).
+  An owning-producer classifier distinguishes `cons`/`reverse`/`concat`/
+  `take`/`drop`/`from_array`/`split_to_seq`/`retain`/`empty` (owned) from
+  `string_seq_tail` (borrowed). `test_string_seq`: 5003 leaks → 0.
+
+### Fixed
+
+- **aetherc double-free crash on the `*StringSeq` tracking fields**
+  (`compiler/codegen/codegen.c`). The `CodeGenerator` is field-initialised
+  (not `calloc`'d); the new seq registry fields were left uninitialised,
+  so the first `clear_seq_vars()` freed garbage pointers — a glibc abort
+  on Linux (`aetherc` SIGABRT/SIGSEGV; 335 programs "Compilation failed")
+  that macOS's allocator silently tolerated. Found with an
+  AddressSanitizer-instrumented `aetherc`; root-caused to a missing field
+  init.
+- **`defer string.release(X)` double-free / use-after-free on magic
+  AetherStrings** (`compiler/codegen/codegen_expr.c`). `string.from_int`/
+  `from_long`/`new` return a refcounted AetherString that the heap-string
+  tracker manages, so the value was freed twice — once by the explicit
+  `string_release`, once by the function-exit defer-free. `string.release(X)`
+  now lowers through the same flag-guarded form as the bare `release()`
+  builtin for any heap-tracked local, freeing once and clearing the
+  tracker. Verified leak-free under AddressSanitizer.
+- **Heap-string leaks from `release()`, `print`/`println` arguments, and
+  `==`/`!=` operands** (`compiler/codegen/codegen_expr.c`,
+  `codegen_stmt.c`). `release()` now reclaims plain malloc'd `char*`
+  results (string.concat/substring/to_upper/to_lower/trim) via a
+  flag-guarded free; the print family and `release`/`string_release` no
+  longer mark their argument escaped (they provably never retain it), so
+  loop accumulators printed/released each iteration are reclaimed; and a
+  heap-returning string expression used directly as a comparison operand
+  is drained after the compare (the binary-operator analogue of the
+  call-argument drain). `test_release_builtin` 102→0,
+  `test_string_leak_loop` 1000→0, `test_fs_realpath` 30→0.
+- **macOS `leaks(1)` gate could hang the CI job indefinitely**
+  (`tests/run_macos_leaks.sh`). Two compounding causes: (1) no per-step
+  timeout, and (2) the leaks output was captured with a `$(...)` pipe — a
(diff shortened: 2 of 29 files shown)


paul-hammant:
MErged PR #635 affects this PR minimally: 
 
The borrowed-pointer warning of #635 on string_array_get (saying "use string.concat(v, "") to make an owned copy or use string_split_to_seq") becomes more load-bearing once the deterministic *StringSeq ownership of #622 lands — string_split_to_seq is now an even better path because the resulting cells auto-free at scope exit. When merged you may want to  strengthen the doc to say "*StringSeq is now the recommended shape; it's reclaimed deterministically." Optional polish, not a conflict.

paul-hammant:
Cross-cutting note worth flagging vs #638 that's landed in o/m first
  
The std.url module uses string_list_new / string_list_add / string_list_size / string_list_get from std.collections. #622 branch rewrites the string_list (std/collections/aether_stringlist.c) to be caps-aware. The public API doesn't change — only internal allocation accounting — so my std.url keeps working unchanged after a merge.  Worth a smoke-test of tests/regression/test_url.ae after merging the two, but no expected breakage.

Issue #698, opened 2026-06-12
Silent narrowing via inference: `x = 0` infers 32-bit int, later `x = some_long` truncates without diagnostic


paul-hammant:
## Summary

A local declared with an untyped integer literal infers 32-bit `int`. A later assignment of a `long` (int64) value to that local silently truncates to 32 bits — no error, no warning. The generated C compiles cleanly because the narrowing is an ordinary implicit conversion there.

```aether
parsed = 0                       // infers int (32-bit)
...
long signed_v = v
parsed = -signed_v               // silently truncates to 32 bits
...
if parsed == target { ... }      // target is long; comparison can never
                                 // match values beyond 32 bits
```

The fix was a one-word change — `long parsed = 0` — which makes the bare spelling a serious foot-gun: the annotation is load-bearing and nothing tells you when you forget it.

## Real-world impact (aedis, the Redis port)

This was root cause #2 of silent data corruption in aedis's listpack codec (sibling of #697, found in the same debugging session):

- `ae_lpStringEqualsInt64` parsed a sought hash-field key into `parsed = 0` (inferred int). For any int-encoded field beyond 32 bits, the parsed value truncated, the equality against the decoded 64-bit entry never matched, and `lpFind` returned nil.
- Observable behavior: `HGET` on a hash with field `-6288338901` returned nil even though `HSET` succeeded. A binary fuzz harness hit 925 failures across 400 trials from this plus #697 combined; 0 after both fixes.
- The trap is camouflaged by inconsistency: the *encode-side twin* of the same parse loop, in the same file, correctly wrote `long parsed = 0` — so the codebase looked like it had the idiom down.

Fix commit in aedis: `ee52fce8e` ("listpack codec: fix two silent data-corruption bugs").

## Wish

Any of, in descending preference:

1. **Error (or warn) on narrowing assignment to an inferred-type local** — if a local's type was inferred from its initializer and a later assignment in the same scope would narrow (long → int), diagnose it. The user can silence it with an explicit `int` annotation if truncation is intended.
2. **Infer from all assignments in the block** — type the local as the widest integer type assigned to it anywhere in its scope (flow-insensitive union), so `parsed = 0` followed by `parsed = some_long` infers `long`.
3. At minimum, a lint flagging `<ident> = <int-literal>` declarations whose local is later assigned a `long`-typed expression.

Related: #697 (int-shift sign-extension) — the other half of the same incident. Together they argue for a small "integer-width hazards" diagnostics pass in the typechecker.

Issue #697, opened 2026-06-12
Int-shift sign-extension: `get_byte(p,4) << 24` lowers to a C int shift and corrupts uint64 expressions


paul-hammant:
## Summary

Shifting an untyped/int-typed operand (e.g. the `int` return of a `std.mem` byte accessor) by 24 or more lowers to a plain C `int` shift. When the resulting value is then combined into a `uint64` expression, C's integer promotion sign-extends it — silently setting all high 32 bits.

```aether
uint64 u = get_byte(p, 1)
u = u | (get_byte(p, 2) << 8)
u = u | (get_byte(p, 3) << 16)
u = u | (get_byte(p, 4) << 24)   // byte >= 0x80: int32 sign-bit overflow,
                                 // then sign-extension on promotion to uint64
```

Generated C:

```c
u = (u | (mem_get_byte(p, 4) << 24));   /* int shift; 0x86 << 24 is negative;
                                           | against uint64_t sign-extends */
```

For a byte of `0x86`, the expression contributes `0xFFFFFFFF86000000` instead of `0x86000000`.

Shift counts >= 32 on an int operand are outright UB — gcc masks the count to 5 bits, so `byte << 48` lands the byte at bit 16:

```aether
b = b | (aether_mem_get_byte_sz(input, index + 6) << 48)   // actually << 16 at runtime
```

## Real-world impact (aedis, the Redis port)

This was the root cause of **silent data corruption** in aedis's listpack codec:

- Negative values in 32-bit listpack encoding were misread: `-2036613105` stored, `-6331580401` read back (high 32 bits polluted by sign-extension, then mangled by the signed-decode path).
- `hashTypeConvertListpack` panicked with **"Listpack corruption detected"**, crashing the server during the replication test suite.
- The same pattern existed (as latent UB) in aedis's `siphash.ae` tail-byte loads (`<< 48`/`<< 40`/`<< 32`/`<< 24`) and `ae_lpDecodeBacklen` (`<< 28`).

The hazard is insidious because the same file is usually *mostly* right: the 64-bit decode path hoisted bytes into `uint64` locals (correct), while the 16/24/32-bit paths inlined the shifts (broken). Nothing warns; tests pass until a byte >= 0x80 reaches the 24+ shift.

## Workaround

Hoist every byte into a `uint64` local before shifting:

```aether
uint64 h4 = get_byte(p, 4)
u = u | (h4 << 24)
```

## Wish

Either of:

[...]

PR #699, opened 2026-06-12
docs: aedis batch ask — mutable module globals + C macro-constant import

paul-hammant:
Downstream spec doc from the Redis port (aedis), following the root-level downstream-report convention and the priority framing of `docs/redis-porting-language-gaps.md`.

Contents:

- **Ask 1 — mutable module-level globals** (`var x: uint64 = ...` at module scope → file-scope C static). Census: deletes `rand.c`, `mt19937-64.c`, and the last buffer edge in `release.c`. Explicitly scoped to module-private statics; C-owned global *import* (the `server.*` spine gate) is flagged as a separate, larger design round.
- **Ask 2 — C macro-constant import** (`extern const EAGAIN: int @c_import` → emits the macro name verbatim; including TU's headers remain the source of truth). Census: un-pins `syncio.ae` from Linux, deletes `release.c`'s string-macro accessors, pre-empts per-constant C accessors across the whole command tier. Object-like macros only.
- **Ask 3 — diagnostics**: pointers to #697 (int-shift sign-extension) and #698 (inference narrowing), both of which shipped as silent data corruption in aedis's listpack codec before being caught by fuzzing.
- Sequenced follow-on list for after this batch.

### DIFF
--- a/aedis-batch-globals-and-macro-import.md
+++ b/aedis-batch-globals-and-macro-import.md
@@ -1,0 +1,178 @@
+# aedis batch ask: mutable module globals + C macro-constant import (+ two integer-width diagnostics)
+
+**Reporter:** aedis (the Redis → Aether port, sibling repo `aedis/`).
+**Toolchain:** `ae 0.239.0`, aetherc at `8d60111`.
+**Platform:** Linux x86_64 (port targets all Redis platforms eventually).
+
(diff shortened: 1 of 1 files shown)


paul-hammant:
Withdrawn — aedis-side convention is untracked root-level report docs, not committed branches. Doc lives in the aether checkout as an untracked file.

PR #748, opened 2026-06-16
std.cryptography wrappers + parser long-long extern spelling

paul-hammant:
Two independent slices, separately committed.

## Summary

1. **std.cryptography: `random_hex(n)` / `random_base64(n)`** — two thin Aether-side wrappers over the existing OS-CSPRNG `random_bytes`, returning a lowercase-hex (2n chars) or RFC 4648 §4 unpadded-base64 string. Closes the third bullet of the in-tree CSPRNG ask (`stdlib-csprng-secure-random-ask.md` from fbs-core); items 1 + 2 (`random_bytes` + UUIDv4) already landed at 0.213.x. Composes existing primitives — no new C, no new externs.

2. **parser: accept `long long` as a two-token extern type spelling** — when the parser sees a second `long` (also lexed as `TOKEN_INT64`) after the first, both are consumed and the resulting `Type` carries `c_alias = "long long"`. The underlying `TypeKind` stays `TYPE_INT64`, so arithmetic and typechecking are unchanged — only the emitted C declaration text differs. Closes the *"Minor, real, cheap"* item from `aedis-core-floor-feature-asks.md`; the larger four asks in that doc become their own follow-up issue (filed alongside this PR).

## Test plan

- [x] `bash tests/integration/cryptography_random_hex/test_cryptography_random_hex.sh` — 6 cases (shape, alphabet, uniqueness, n=0, length-correlation)
- [x] `bash tests/integration/long_long_extern/test_long_long_extern.sh` — 4 cases (single arg+return, large value, mixed `long long` ↔ `int64_t`)
- [x] `bash tests/integration/cryptography_v2/test_cryptography_v2.sh` — unaffected, 8 cases still green
- [x] `bash tests/integration/cryptography_sha/test_cryptography_sha.sh` — unaffected, 6 cases still green
- [x] `make ci` — 665/668 green. The 3 failures are pre-existing HTTP h2 / proxy / port-binding flakes (see memory `[Windows HTTP test flakes]`); each passes on retry. Neither slice touches HTTP or scheduler code.


paul-hammant:
Follow-up issue #749 filed for the three remaining aedis P0/P1 compiler asks (fn-ptr struct fields + params, restricted varargs call-through, longdouble) — with probe insights from this session on where the existing infrastructure already covers more than the ask framing suggests.

PR #830, opened 2026-06-20
test(c_import): regression guard for struct-without-typedef emission

paul-hammant:
The fix that makes `aetherc` emit `struct Name *` (rather than bare `Name *`) for pointers to `@c_import` structs **landed on main via #534**, but its regression test never did — so there's currently no guard against that codegen path regressing.

This is a **test-only** PR adding that guard:

- `widget.h` — a header that deliberately ships **no** `typedef struct widget widget;`, mirroring POSIX's `struct tm` (`<time.h>`), `struct stat` (`<sys/stat.h>`), `struct sockaddr` (`<sys/socket.h>`), etc.
- `probe.ae` — `@c_import`s that struct, casts a raw `ptr` to `*widget`, reads fields, prints them.
- The `.sh` driver asserts the generated C contains `struct widget` (and emits **no** typedef for it), compiles + links against the bare header, runs, and produces the expected output.

Without the fix, the generated C fails with *"unknown type name 'widget'"*. Verified **PASS** against current main (the fix is present; only the test was missing).

No production code changes — this just plugs a coverage gap left when the original `feat/std-mem-byte-sz` branch's code half merged (squashed) without its test half.

### DIFF
--- a/tests/integration/c_import_struct_no_typedef/aether.toml
+++ b/tests/integration/c_import_struct_no_typedef/aether.toml
@@ -1,0 +1,11 @@
+[[bin]]
+name = "probe"
+path = "probe.ae"
+
+[build]
+# Force-include the C header that owns `struct widget` into every TU,
+# including the Aether-generated .gen.c.  The header deliberately does
+# NOT carry a `typedef struct widget widget;`, so the only way this
+# compiles is if aetherc emits `struct widget *` (not bare `widget *`)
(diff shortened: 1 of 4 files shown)


paul-hammant:
Closing to conserve CI minutes — folding this test into a larger bc-csharp crypto tranche so it rides one CI run. Will re-add the c_import guard there.

PR #857, opened 2026-06-22
docs: tighten README Core Features + accuracy pass across user-facing docs

nicolas-maman:
## Summary

Two related documentation improvements, in separate commits:

1. **README — tighten Core Features.** The list had grown into paragraph-length bullets that duplicated the sections right below them (Runtime Features, Capabilities & Sandboxing) and the deep-dive docs. Each is now a single concrete, scannable line with a link to its deep-dive. No claim was dropped that isn't covered downstream, and every retained claim was checked against the code.

2. **Docs — accuracy pass.** Audited the user-facing docs against the actual compiler/runtime/stdlib and fixed confirmed hallucinations and stale claims. **Every code example changed was compile-checked with the in-tree compiler.**

## Headline fixes

**Code examples that did not compile (now do):**
- `tutorial.md` / `getting-started.md` — pattern-match clauses had a trailing `;` the parser rejects.
- `tutorials/02` — the "functional style" recursive examples coerced the list parameter to a fixed-size array (invalid C); rewritten on the real `*StringSeq` head/tail type.
- `type-annotation-style-guide.md` — struct-typed prefix params (`Point p`) and `state x: int` don't parse; switched to the forms that do.
- `type-inference-guide.md` — `long 0` isn't value syntax.
- `http-server.md` — `http.http_server_*` examples don't resolve (the `http.` namespace already consumes the `http_` prefix); the request-hook isn't actually exported.
- `language-reference.md` — `as` also accepts `fn(...) -> R` and `T[]`; the glob-import example used a module that doesn't ship.

**Stale API shapes / wrong claims:**
- `std.json` getters and `stringify` return `(value, err)` tuples now.
- `std.cryptography` ships HMAC-SHA256, MD4/MD5, and a streaming digest API (moved out of "not supported"); removed a phantom "RSA" from the std table (public-key crypto is in `contrib`).
- All four `std.fs` structured-error wrappers (copy/move/realpath/chmod) ship.
- `memory-management.md` — actors use one NUMA-aware allocation (not an arena); `defer` is block-scoped and fires **per loop iteration** (verified by emitted C); function-exit free emits `aether_heap_str_free`; the `@retain` carriers were wrong (the real ones are `string_retain`/`release`/`free` + `map_put_string_owned`'s key — `string_list_add`/`set` deliberately copy and are **not** retainers).
- `containment-sandbox.md` — `bind()`/`accept()` ARE intercepted; the JS host flag is `AETHER_HAS_DUKTAPE`; the shared-map example now uses a host that exports it.
- `cic-help.md` — `ae help` is **implemented and shipped** (`tools/ae_help.c`), not "proposed"; reframed to match what was actually built (it runs `aetherc` as a subprocess and parses its diagnostics, rather than hooking into the typer as the original sketch planned).
- `tcp.send` → `tcp.write`; assorted version and line/symbol-reference drift; `--sandboxed-build` and `@derive(format/clone/hash)` flagged as not-yet-implemented.

## Method

The audit and fixes were fanned out across subagents; each fix-agent owned a disjoint set of docs, re-verified every claim against the source, and compile-checked every code example it touched. `@derive` is a good example of the bar: a parser comment implied `eq/format/clone/hash`, but `compiler/analysis/derive.c` only implements `eq` — so the docs (and README) keep `@derive(eq)`.

## Scope / follow-up

This pass covers the **user-facing** docs (the README and ~20 docs people read to *use* Aether). Not yet audited: design/rationale/proposal docs, the `cross-references/` and `notes/` material, porting notes, and the auto-generated `docs/api/*.html`. Happy to do a second pass over those if you want.


nicolasmd87:

PR #858, opened 2026-06-23
docs(api): generate stdlib API docs from .ae exports, not C headers

nicolas-maman:
## Summary

The generated stdlib API docs (`docs/api/*.html`) were produced by a generator (`tools/docgen`) that parsed **C headers**, so they documented the wrong layer: `*_raw` internals and header-only utilities, **wrong signatures** (e.g. json `get_string` shown as a single-value C function instead of the real `(string, string)` tuple wrapper), and they **omitted most of the public API** (os/fs/io/string were each missing dozens of functions). The HTML was also unversioned and ~2 months stale.

This rewrites the generator to document the **user-facing `.ae` API**: for each `std/<module>/module.ae`, it parses the `exports (...)` list and renders each exported function's parameters, return type (including tuple returns), kind (fn/extern/const/struct), and doc-comment. The HTML emitter/templates are preserved. **Coverage goes from 9 modules to 47** (every std module with an exports block), regenerated via `make docs`.

## Bugs found and fixed while reviewing the rewrite

- **Stack overflow in `strip_annotations()`** — it wrote up to 1024 bytes into 256/512-byte caller buffers. This silently truncated every multi-parameter signature (`object_get(obj)` instead of `object_get(obj, key)`) and aborted outright in isolation. Now bounded by the destination capacity.
- **Call-site names** — the `<module>.` namespace consumes a leading `<module>_` prefix, so `string_pad_start` is now documented as `pad_start` (called as `string.pad_start`); exports without the prefix (json's `get_string`) are unchanged.
- **Annotation-strip spacing** — dropped the stray space left before a separator (`(string @heap, int)` → `(string, int)`).

## Verification

- `tools/docgen` builds `-Wall -Wextra -Werror` clean.
- Regenerated output spot-checked: json getters show their `(value, err)` tuple returns; `object_get(obj, key)`/`object_set(obj, key, value)` render all params; fs/os/io/string list their full export sets; no `@heap`-spacing artifacts (0 remaining); per-module function counts are sane with no prefix-collision collapse.

Note: the diff is large because it includes the regenerated HTML for all 47 modules. The meaningful change to review is `tools/docgen/docgen.c`.

### DIFF
--- a/docs/api/actors.html
+++ b/docs/api/actors.html
@@ -1,0 +1,156 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>actors - Aether</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+<body>
+<nav class="sidebar">
+  <div class="sidebar-header">
+    <h1><a href="index.html">Aether</a></h1>
+    <p class="tagline">Standard Library</p>
+  </div>
+  <input type="text" id="search" placeholder="Search actors..." onkeyup="searchModule()">
+  <h2>Exports</h2>
+  <ul class="function-nav">
+    <li><a href="#aether_actor_register">aether_actor_register</a></li>
+    <li><a href="#aether_actor_whereis">aether_actor_whereis</a></li>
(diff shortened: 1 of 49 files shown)


nicolasmd87:

Issue #701, opened 2026-06-12
Mutable module-level globals (var at module scope)


paul-hammant:
**Reporter:** aedis (Redis → Aether port, sibling repo).
**Toolchain:** \`ae 0.239.0\`, aetherc at \`8d60111\`.
**Platform:** Linux x86_64 (port targets all Redis platforms eventually).

This is one of two leaf-tier asks from the aedis side that together delete most of the remaining residual C in the port. See also the macro-import companion ask and the two integer-width diagnostics (#697, #698).

## Problem

Aether has no mutable state at module scope. Any C function whose whole job is \"algorithm + one persistent word of state\" therefore keeps a vestigial C file holding the state plus get/set accessors, with the real logic already in \`.ae\`.

## Call-site census (aedis today)

| File | Residual C | State |
| --- | --- | --- |
| \`rand.c\` | 22 lines | one \`static uint64_t rand48_x\` + 2 accessors; the entire 48-bit LCG already lives in \`rand.ae\` |
| \`mt19937-64.c\` | whole file kept | the 312-word Mersenne Twister state array + \`mti\` index |
| \`release.c\` | 1 of its 4 residual lines | a 32-byte static buffer caching the formatted build id (formatting + cache policy already in \`release.ae\`) |

Same shape is guaranteed to recur in every module with a \`static\` cache or PRNG: \`dict.c\`'s resize policy flags, \`lru_clock\` style counters, etc.

## Proposed spelling

\`\`\`aether
var rand48_x: uint64 = 0x1234ABCD330E      // module scope, mutable

redisLrand48() -> int32_t {
    rand48_x = (rand48_x * 0x5DEECE66D + 0xB) & 0xFFFFFFFFFFFF
    return (rand48_x >> 17) & 0x7FFFFFFF
}
\`\`\`

## Required semantics

- Module-scope \`var name: type = const-expr\` emits a file-scope \`static <ctype> name = <value>;\` in the generated TU.
- Reads/writes from functions in the same module emit plain C identifier access. No accessor indirection, no TLS unless asked.
- Initializer restricted to constant expressions (no static constructors needed — matches C).
- Cross-module visibility can be punted: module-private (\`static\`) is enough for every aedis call site above. \`pub var\` / extern-linkage globals can come later if ever needed.
- Threading: plain non-atomic statics, same as the C they replace. Redis guards these by design (main-thread-only access); aedis does not ask for atomics here.

## What it deletes

[...]

PR #859, opened 2026-06-23
fix(codegen): module-global string var writes the static, not a local shadow

nicolas-maman:
## Summary

A **string-typed module-level `var` was miscompiled**: a setter's write went to a function-local shadow instead of the module global, so cross-function reads never saw it. Found while updating `docs/per-process-config.md` (it had documented module-level `var` as unsupported; once documented, the string case turned out to be broken).

Scalar module globals (`int`/`long`/`uint64`/`ptr`, #701) worked; only the string-typed case hit this path.

## Root cause

The heap-string tracker's `collect_heap_string_var_names` treated the string global as a heap-tracked local, so codegen:
1. hoisted a function-local `const char* <name> = NULL;` shadow over the file-scope `static`,
2. routed the setter's write through the reassignment wrapper into that **shadow**, and
3. pushed a function-exit defer-free over a **process-lifetime global**.

So `set_token("x"); ... get_token()` returned the initializer, not `"x"`.

```c
// before (buggy)                          // after (fixed)
static const char* token = "none";         static const char* token = "none";
void set_token(const char* s) {            void set_token(const char* s) {
    const char* token = NULL;  // shadow       token = s;   // writes the static
    { ...; token = s; ... }
    if (_heap_token) free(token); // !!
}
```

## Fix

`collect_heap_string_var_names` skips module globals — mirroring the scalar #701/#744 hoist skips — so the write routes to the file-scope `static` (a real lvalue), with no shadow and **no defer-free of the global** (so no UAF/double-free; the latest value persists for the process lifetime, exactly like the scalar model).

Secondary: a write-only setter (`set_x(v) { x = v }`) of a module global produced a spurious `warning[W1001]: unused variable 'x'` — the bare `x = expr` parses as a declaration whose LHS the reference pass skips. `typecheck_program` now collects the module-global `var` names and threads them through `check_unused_variables` so those writes aren't mistaken for unused locals. (The scalar case escaped this only because its setters read-modify-write.)

## Test

Adds `tests/integration/module_globals/string_setter.ae`: cross-call reads of a string module global through a setter (initial value, two reassignments), plus a scalar global in the same program guarding against a #701 regression.

## Verification

- `make compiler ae stdlib EXTRA_CFLAGS="-Werror"` — clean, zero warnings.
- New test passes; existing `module_globals` tests (counters/lcg_prng/nested_block_init) stay green.
- `make examples` — 83/83.
- `make test-ae` — 739/739.
- The W1001 false-positive is gone.

### DIFF
--- a/compiler/analysis/typechecker.c
+++ b/compiler/analysis/typechecker.c
@@ -1648,13 +1648,37 @@ static void collect_references(ASTNode* node, TrackedVar* vars, int var_count) {
     }
 }
 
+/* Is `name` a module-level `var` global? Such a name, when it appears
+ * as the LHS of a bare `name = expr` inside a function, parses as an
+ * AST_VARIABLE_DECLARATION but is NOT a new local — it is a WRITE to
+ * the file-scope `static` (#701). Treating it as a local declaration
+ * makes the unused-variable pass flag pure-setter functions
+ * (`set_x(v) { x = v }`) with a spurious "unused variable 'x'": the
+ * write-only assignment never reads the name, and collect_references
+ * deliberately skips a declaration's LHS. The scalar case escaped
+ * notice only because its setters read-modify-write (`x = x + 1`).
+ * Caller passes the module-global name set so these writes are not
+ * mistaken for unused local declarations. */
+static int is_module_global_name(const char** global_names, int global_count,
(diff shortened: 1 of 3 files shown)


nicolasmd87:

Issue #1035, opened 2026-07-06
std.os argv API: documented qualified form os.aether_args_count() fails to resolve; only the unqualified spelling works


nicolas-maman:
**Version:** ae 0.358.0, commit `9b2a860`. Found wiring `--bench` argv parsing into a game executable. Related: same session as #1025, #1033.

## Symptom

The language reference documents qualified argv access (docs/language-reference.md, ~line 384):

```aether
n = os.aether_args_count()
println("argv[${i}] = ${os.aether_args_get(i)}")
```

That exact form fails to resolve:

```
error[E0301]: Undefined function 'os.aether_args_count'
```

The unqualified spelling compiles and runs:

```aether
import std.os
main() { println("n=${aether_args_count()}") }   // works
```

## Root cause (as far as I traced it)

`std/os/module.ae` exports `aether_args_count`, `aether_args_get`, `aether_argv0` alongside the `os_`-prefixed API. Qualified resolution appears to work by stripping the module prefix from exported names (`os_getenv` → `os.getenv` resolves; verified). Exports that don't carry the `os_` prefix fall outside that rule, so there is *no* qualified spelling that reaches them — `os.aether_args_count`, `os.args_count`, and `os.count` all fail — while the bare extern name leaks through unqualified. Same likely applies to the other non-prefixed exports in that list (`args_seal`, `args_sealed`, `exit`, …) — I only verified the argv trio.

Repro matrix (all with `import std.os`):

| call | result |
|---|---|
| `os.getenv("HOME")` | ✅ resolves (prefix-stripped from `os_getenv`) |
| `os.aether_args_count()` | ❌ E0301 — the documented form |
| `os.args_count()` | ❌ E0301 |
| `aether_args_count()` | ✅ works, unqualified |

## Why it's worth fixing rather than just re-documenting

The unqualified escape hatch undermines the module qualification story: a symbol you imported from `std.os` is reachable without any namespace at all, which collides with the `exports`-visibility model everywhere else. Two consistent options:

1. **Resolver:** make non-prefixed exports reachable under the module name (`os.aether_args_count`), matching the docs as written; ideally alias the ergonomic `os.args_count` / `os.args_get` / `os.argv0` while at it.
2. **Docs + module:** rename the exports to `os_args_count` etc. so prefix stripping yields `os.args_count`, and fix language-reference.md:384 and stdlib-api.md:338-340.

[...]

Issue #1033, opened 2026-07-06
FFI: by-value struct returns work (tuple externs, #271) but by-value struct arguments don't — the remaining half of raylib-class interop


nicolas-maman:
**Version:** ae 0.358.0. Found building CARDFALL's raylib-backed renderer in pure Aether. Related: #271 (built the return half), #266, #746, #1025 (same effort).

## What already works — verified, and worth advertising

The tuple-return extern mechanism (docs/c-interop.md §120, from #271) is in practice **by-value struct return** for any C struct whose layout matches the synthesized `_tuple_*` typedef. raylib's `Image LoadImage(const char*)` — layout `{void*, int, int, int, int}` — binds today with zero glue:

```aether
@extern("LoadImage") load_image(path: string) -> (ptr, int, int, int, int)

main() {
    data, w, h, mips, fmt = load_image("frame.png")
    // -> Image: 400x420 mips=1 fmt=7   (verified against a real PNG)
}
```

This is a genuinely good pattern and I couldn't find it documented as such — c-interop.md presents tuple returns as "C functions that return multiple logical values", not as "this is how you bind a struct-returning C function". A short subsection with the raylib example would save every FFI user a shim.

## The gap: parameter position

There is no equivalent for by-value struct **arguments**, and that's most of any real C API. raylib again:

```c
void ImageDrawTriangle(Image *dst, Vector2 v1, Vector2 v2, Vector2 v3, Color color);
```

Not declarable. Today this forces a hand-written flat-scalar shim per function:

```c
void rl_img_triangle(void* p, double x1, double y1, double x2, double y2,
                     double x3, double y3, int r, int g, int b, int a) {
    ImageDrawTriangle((Image*)p, (Vector2){(float)x1,(float)y1}, ...,
                      (Color){r,g,b,a});
}
```

It works — CARDFALL's whole renderer runs on a 120-line shim — but it's boilerplate that scales with the API surface, argument packs like `r,g,b,a` erase the type, and it's an extra call frame on the hottest path in a per-pixel rasterizer.

## Proposal: mirror #271 into parameter position

The codegen already synthesizes C struct typedefs for tuple returns and the ABI already matches Aether's own multi-return functions. The symmetric move is tuple-typed extern parameters, lowering to the same synthesized typedefs, called with existing tuple values or inline groupings:

```aether
[...]

PR #1039, opened 2026-07-07
fix(ae): AETHER_CACHE_DIR override + atomic temp-then-rename cache publish

### LINKED ISSUE #1032: ae cache: AETHER_CACHE_DIR override + atomic temp-then-rename publish (concurrent invocations race on shared cache slots)
## Two related defects in the `~/.aether/cache` design

Surfaced when a sandboxed agent (read-only `$HOME`, writable workspace) tried to redirect the cache and found it couldn't; investigating that also showed the cache is not safe under concurrent invocations. Multiple agents/terminals now routinely share one `$HOME` on a dev box, so both are practical, not theoretical.

### 1. The cache location is hard-wired — `AETHER_HOME` does not move it

`init_cache_dir()` (tools/ae.c:358) builds `$HOME/.aether/cache` from `HOME`/`USERPROFILE` only. `AETHER_HOME` is consulted solely for toolchain-root resolution, so a runner whose `$HOME` is read-only (agent sandboxes, hermetic CI) has no way to redirect build artifacts short of overriding `HOME` itself, which drags everything else with it. Observed in the wild:

```
/usr/bin/ld: cannot open output file /home/paul/.aether/cache/b4743634eb219074: Read-only file system
```
(after the runner had already tried `AETHER_HOME=/tmp/...` and reasonably expected it to help)

**Proposed:** a per-process override with a simple precedence chain in `init_cache_dir()`:

1. `AETHER_CACHE_DIR` set → use verbatim
2. else → today's default `$HOME/.aether/cache` (`USERPROFILE` on Windows)

Deliberately **no** fallback to `$AETHER_HOME/cache`: `AETHER_HOME` means "where the (often read-only) toolchain lives"; overloading it to also mean "where writable artifacts go" conflates the two — it is exactly the read-only location the sandboxed runner was trying to escape. Two variables, two meanings.

### 2. Concurrent invocations race on shared cache slots

Three defects compose into an exec-of-truncated-binary window:

- **`ae run` links directly into the final cache slot.** On a miss, `exe_file` *is* `~/.aether/cache/<key>` — "write exe directly to cache slot (no extra copy needed)" (tools/ae.c:2944-2951). While gcc/ld writes, the slot exists and is partial.
- **The hit path is a TOCTOU against that:** `path_exists(cached_exe)` → exec (tools/ae.c:2926-2937). A second invocation with the same key landing mid-link execs a half-written binary — ENOEXEC at best, a garbage segfault (indistinguishable from a program bug) at worst.
- **`ae build` populates the cache with a non-atomic `copy_file`** into the final slot (tools/ae.c:5206-5212) — same partial-read window for any concurrent `ae run` of the same key.

Why it hasn't visibly bitten: collisions need the *same* cache key (same source + flags + deps hash), so parallel runners over different `.ae` files occupy disjoint slots, and on Linux relinking a binary that another process is currently *executing* fails with ETXTBSY rather than corrupting it. The dangerous window is same-key concurrency during the build/copy phase — two agents, a re-run racing a still-linking first run, a `-j` build invoking the same probe twice.

**Proposed:** standard atomic publish. Link/copy to `<slot>.tmp.<pid>` **in the same directory**, then `rename()` onto the slot — atomic on POSIX, last-writer-wins with every published file complete; readers either see the old complete file, the new complete file, or a miss. (Windows: `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` semantics via `_wrename`-equivalent; same-directory keeps it same-volume.) Clean up orphaned `.tmp.*` older than some horizon while at it.

The two changes compose: `AETHER_CACHE_DIR` solves "this process needs a different cache location"; atomic publish solves "processes sharing a location must never observe partial slots."

### Test plan

- Unit-ish: `AETHER_CACHE_DIR=$(mktemp -d) ae run …` populates the override dir and leaves `~/.aether/cache` untouched; unset falls back to the default.
- Concurrency hammer: N parallel `ae run`s of the same freshly-generated source (guaranteed shared key, cold cache), assert every invocation exits with the program's own status — no ENOEXEC, no signal deaths, and the surviving slot is a valid executable.
- Read-only-home simulation: `HOME=<ro dir> AETHER_CACHE_DIR=<rw dir> ae run …` succeeds (the observed sandbox scenario).

_Found via a Codex session whose sandbox mounts `$HOME` read-only; its `AETHER_HOME=/tmp/...` attempt not redirecting the cache is what exposed defect 1, and the code reading for that exposed defect 2._

paul-hammant:
Closes #1032.

## AETHER_CACHE_DIR

`init_cache_dir()` now honors a per-process `AETHER_CACHE_DIR` override, falling back to the existing `$HOME/.aether/cache` default. As argued in the issue, `AETHER_HOME` deliberately still does not move the cache — it names the (often read-only) toolchain root, while this is a writable artifact directory. The reported sandbox scenario (`$HOME` read-only, workspace writable) now has a one-variable fix.

## Atomic publish

Both cache writers now produce `<slot>.tmp.<pid>` in the cache directory and publish with an atomic rename (POSIX `rename(2)`; `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` on Windows):

- `ae run` no longer points the linker at the final slot — it links to the private temp and renames on success. On rename failure (exotic fs) it degrades to running the temp as an uncached build.
- `ae build` no longer `copy_file`s directly onto the slot — copy to temp, then rename.

The `path_exists → exec` hit path is unchanged and now safe: readers see the old complete file, the new complete file, or a miss — never a partial slot. Orphaned `.tmp.<pid>` files from killed writers are swept when older than an hour (age-gated so an actively-linking neighbor is never reaped), once per process from `init_cache_dir()`.

## Tests

- `tests/integration/cache_dir_override/` — override populates only the override dir; second run hits from it; and the exact reported scenario: `HOME=<chmod 555 dir> AETHER_CACHE_DIR=<rw dir> ae run` succeeds.
- `tests/integration/cache_concurrent_publish/` — 8 parallel `ae run`s of the same freshly-generated source (shared key, guaranteed-cold private cache): all must exit 0 with correct output, and no `.tmp.*` orphans may survive. Pre-fix this races the linker against exec of the half-written slot.

Both pass locally; `.sh` integration tests are auto-discovered by the Makefile runner.

CHANGELOG entry included under `## [current]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the
 next version number before tagging the release.
 
+## [current]
+
+### Fixed
+
+- **`ae` exe cache: `AETHER_CACHE_DIR` override + crash-proof concurrent
+  publishing** (#1032). The cache location was hard-wired to
+  `$HOME/.aether/cache`, unusable for runners with a read-only `$HOME`
+  (agent sandboxes, hermetic CI) — `AETHER_CACHE_DIR` now redirects it
+  per-process (`AETHER_HOME` deliberately still doesn't: toolchain root
+  and artifact dir are different concepts). Concurrent same-key
+  invocations also raced on shared slots: `ae run` pointed the *linker*
+  at the final slot and the hit path was exists→exec, so a second
+  invocation landing mid-link exec'd a truncated binary; `ae build`
+  populated the slot with a non-atomic copy. Both writers now produce
+  `<slot>.tmp.<pid>` and publish with an atomic rename (`MoveFileEx` on
+  Windows), so readers see a complete file or a miss — never a partial.
+  Orphaned temps from killed writers are swept after an hour. Two new
+  integration tests: the read-only-`$HOME` override scenario, and an
+  8-way parallel cold-cache hammer.
+
 ## [0.363.0]
 
 ### Added
--- a/tests/integration/cache_concurrent_publish/test_cache_concurrent_publish.sh
+++ b/tests/integration/cache_concurrent_publish/test_cache_concurrent_publish.sh
@@ -1,0 +1,72 @@
+#!/bin/sh
+# Regression for #1032 (half 2): concurrent same-key `ae run`s must never
+# exec a partially-written cache slot.
+#
+# Before the fix, `ae run` pointed the LINKER at the final cache slot and
+# the hit path was `path_exists -> exec`, so a second invocation landing
+# mid-link exec'd a truncated binary (ENOEXEC or a garbage segfault). Now
+# writers link to <slot>.tmp.<pid> and publish with an atomic rename, so
+# every reader sees a complete file or a miss.
+#
+# This hammers N parallel runs of the same fresh source (shared key, cold
+# cache) and requires every one to exit 0 with the right output.
+set -u
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+AE="$ROOT/build/ae"
+
+if [ ! -x "$AE" ]; then
+    echo "  [SKIP] $AE not built (run make first)"
+    exit 0
+fi
+
+TMPDIR_T="$(mktemp -d)"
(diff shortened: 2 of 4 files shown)


paul-hammant:
Folding into a single combined PR for #1032 + #1033 + #1035 to conserve Actions minutes — same commits, one CI run.

Issue #1054, opened 2026-07-07
return match {...} as a bare expression emits void return + dead arms (garbage result)


paul-hammant:
## Bug

Returning a `match` expression directly — `return match x { ... }` — miscompiles: the codegen emits a bare `return;` (void) followed by an orphaned match statement whose arm bodies are dead expression-statements. The function returns garbage / crashes.

**Independent of enums** — reproduces with a plain `int` scrutinee.

## Repro

```aether
classify(n: int) -> string {
    return match n {
        0 -> "zero"
        1 -> "one"
        _ -> "many"
    }
}
main() { println(classify(1)) }   // prints garbage, not "one"
```

Emitted C (abridged):
```c
const char* classify(int n) {
    return;                 // <-- void return, wrong
    {
        int _match_val = n;
        if ((_match_val == 0)) { "zero"; }   // <-- dead expr-statement
        else if ((_match_val == 1)) { "one"; }
        else if ((_match_val == 2)) { "many"; }
    }
}
```

## Workaround

Assign then return — works correctly:
```aether
classify(n: int) -> string {
    r = match n { 0 -> "zero"  1 -> "one"  _ -> "many" }
    return r
}
```

## Root cause (guess)

`match` is being lowered as a *statement* in return position rather than as a value-producing expression. The `AST_RETURN_STATEMENT` codegen likely doesn't special-case a `match` (or `if`) expression operand the way an assignment RHS does. The assign-then-return path works because the match-as-initializer path lowers the arms as assignments to the target.

Found while adding enum tests (#1044) — the enum `match` works fine via the assign-then-return idiom; this is a separate, pre-existing gap.

Issue #1047, opened 2026-07-07
Ranged and multi-value match/switch cases (case 1..=5, 20..<30:)


paul-hammant:
_From an Odin→Aether feature comparison. **Tier 1** — cheap, purely additive, obviously useful._

## Gap

Aether's `match` / `switch` arms are literal-only (single int/string literal, wildcard, or — for `match` — list/sum/optional patterns). Range-branching a scalar forces you into guarded function clauses (`when x >= 90 -> ...`), which is far heavier than the task deserves for a simple numeric bucket.

Odin (`examples/demo/demo.odin`) allows ranged and comma-listed cases:

```odin
switch c {
case 'A'..='Z', 'a'..='z', '0'..='9': ...   // ranges + list in one case
case 20..<30:                                 // half-open range
}
```

## Proposed for Aether

```aether
match score {
    90..=100 -> "A"
    80..<90  -> "B"
    70..<80  -> "C"
    _        -> "F"
}

match ch {
    'a'..='z', 'A'..='Z' -> "letter"
    '0'..='9'            -> "digit"
    _                    -> "other"
}
```

- `..=` inclusive, `..<` half-open (matches the existing `for i in 0..5` exclusive-range spelling — keep range operators consistent).
- Comma-separated values/ranges in one arm.
- Applies to both `match` and the C-style `switch`.

## Why it's cheap

- Parser: extend case-label parsing to accept a range and a comma list (the range operators already exist for `for`).
- Typechecker: elements must be the same ordinal type (int/byte/char); overlap detection is a nice-to-have warning, not required for v1.
- Codegen: a ranged int case → `if (x >= lo && x <= hi)` in the lowered branch chain; no new runtime.

No type-system changes, no new AST *types* (just range/list children on the existing case node). Purely additive — existing literal cases are unaffected.

PR #1055, opened 2026-07-07
feat: first-class enum types + enum-backed bit_set (#1044, #1046)

### LINKED ISSUE #1044: First-class enum types (foundation for bit_set, enum-indexed arrays, cleaner C-enum FFI)
_From an Odin→Aether feature comparison. Tier 2, but a **prerequisite** for #bit_set and a dependency several other items want, so raising it first._

## Gap

Aether has no user-facing `enum` type. Enumerations exist only *internally* as the tag of a sum type (`type Shape = Circle | Rect`). There is no named, iterable, switchable enum with explicit integer values — the thing every systems language treats as a primitive.

Odin (`examples/demo/demo.odin`) treats enums as first-class: named type, explicit values, enum-indexed arrays (`[Color]int{.Red = 1, ...}`), `#sparse` variants for gapped enums, and implicit selector syntax (`.Red` inferred from context).

## Why it matters

1. **It's the prerequisite for `bit_set`** (set-of-enum backed by an integer — separate issue).
2. **Enum-indexed arrays** (`[Direction]Vector` / `[State]handler`) are a common, allocation-free lookup table.
3. **Cleaner C-enum FFI** — today a C enum is bound as bare `int` consts; a real enum type would round-trip with names.
4. Enum-keyed `match` with exhaustiveness (which you already do well for sum tags) generalizes naturally.

## Proposed shape (sketch — design discussion welcome)

```aether
enum Direction { North, East, South, West }        // implicit 0..3
enum Errno { Ok = 0, NotFound = 2, Perm = 13 }      // explicit values

d = Direction.North
d = .North                                          // implicit selector where type is known

match d {
    North -> { ... }
    _     -> { ... }
}

// enum-indexed array (allocation-free table)
names: [Direction]string = [.North = "N", .East = "E", .South = "S", .West = "W"]
```

Codegen is a C `enum` (or a plain int with named constants); zero runtime cost. New `TypeKind` (e.g. `TYPE_ENUM`) + AST node; the tag machinery already in the sum-type lowering is a starting point.

## Scope notes / decisions for the design

- Interaction with sum types: keep them distinct (sum = tagged union of *structs*; enum = named integers), or unify? Recommend keeping distinct — they solve different problems.
- Iteration: an `enum_values(Direction)` or `for d in Direction` form is a nice-to-have, not v1-blocking.
- `#sparse`-style gapped enum-indexed arrays can be a follow-up.

## Related

- Blocks the `bit_set` issue.
- Exhaustiveness/narrowing machinery already exists for sum-type `match` (`docs/language-reference.md` §sum types) — reuse it.

_Context: Aether already matches or beats Odin on defer, tuple returns/params, struct-by-value FFI, `when` static-if, sum types + exhaustive `match`, distinct types, `T!`/`expr!` error sugar. This is one of the few genuine primitive gaps._

### LINKED ISSUE #1046: bit_set — set-of-enum backed by an integer (flags/masks without manual bit-twiddling)
_From an Odin→Aether feature comparison. **Tier 1.** Depends on first-class enums (linked)._

## Gap

Bit-flag / permission-mask work in Aether is manual `&` / `|` / `<<` plus `std.bits` helpers (`popcount32`, `rotl32`, ...). There's no set-of-enum type. This shows up constantly in exactly the code Aether cares about — the sandbox/capability layer manipulates permission masks by hand.

Odin's `bit_set[Enum]` (`examples/demo/demo.odin`) is an integer-backed set with delightful ergonomics:

```odin
Days :: distinct bit_set[Day]
WEEKEND :: Days{.Sunday, .Saturday}
if .Sunday in today { ... }
card(today)                 // popcount
a <= b                      // subset test
```

## Proposed for Aether

```aether
enum Perm { Read, Write, Exec }
type Mask = bit_set[Perm]

m: Mask = { .Read, .Write }
m = m + { .Exec }           // union
m = m - { .Write }          // difference
if .Read in m { ... }
n = card(m)                 // popcount
sub = a <= b                // subset
```

Operators: `+ - += -= in not_in <= >= < >` and `card`. Optional explicit backing type (`bit_set[Perm; u16]`) like Odin, for FFI-width control.

## Codegen

Lowers to a plain integer (default width = smallest int holding the enum's cardinality). `{ .Read, .Write }` → `(1<<Read)|(1<<Write)`; `in` → `(m >> e) & 1`; `card` → popcount; set ops → bitwise ops. Zero runtime cost, no allocation, no conflict with any existing model.

## Dependency

Requires first-class `enum` types (the element type of a `bit_set`). Blocked on that issue — the two are a natural pair to land in sequence (enums first).

## Concrete first consumer

The runtime sandbox / capability code (`runtime/sandbox/`, `std.capsicum` R_*/F_* masks) is hand-rolling exactly this pattern today — a `bit_set[CapRight]` would replace the `0x0001`-style const blocks with a typed set.

paul-hammant:
Closes #1044. Closes #1046.

Delivers the tier-1 pairing from the Odin comparison: a real `enum` type usable as a first-class value, plus an integer-backed `bit_set` over it. #1044 (enums) is the substrate #1046 (bit_set) needs, so they land together.

## Enums (#1044)

```aether
enum Perm { Read, Write, Exec }

grant(p: Perm) -> Perm {          // param + return
    if p == .Read { return .Write }  // == and return-position values
    return p
}

struct Acl { owner: Perm }        // struct field
Perm p = .Write                   // variable decl, implicit selector
match p { Write -> "w"  _ -> "?" } // match arm
```

Lowers to a C `int`, zero runtime cost. Implemented via a `resolve_enum_types` typechecker pass that rewrites bare `TYPE_STRUCT{EnumName}` use-site annotations to `TYPE_ENUM` and stamps every `.Member` node with its integer index by name. Stamping in that pass (not a later typecheck route) is what makes indices correct in **every** position — the original commit emitted `return 0` for every member in return position because only some paths set the index.

## bit_set (#1046)

```aether
type Mask = bit_set[Perm]
Mask m = { .Read, .Write }
m += { .Exec }                    // union (also -, +=, -=)
if .Read in m { ... }             // in / not_in
n = card(m)                       // popcount
if rw <= m { ... }                // subset (<, <=, >, >=)
```

Lowers to a `uint64_t`; every op is integer arithmetic (`1ULL << idx`, `__builtin_popcountll`, bitwise). No allocation.

## Correctness (all found by an adversarial review of the initial commit; the single happy-path test caught none of them)

- **Enums weren't actually first-class** — var decls / params / fields / `==` / `match` all failed; only bit-index selectors worked. Fixed (the headline change above).
- **>64-member bit_set was silent UB** (index 64+ shifts past the uint64_t width, also a `-Wshift-count-overflow` that trips `-Werror` on CI) → now a clean compile error, in a pass that's order-independent of enum-vs-alias declaration order.
- **Duplicate enum members** silently accepted → rejected.
- **`card` hijacked any user function named `card`** → now only intercepts a bit_set argument.

## Scope / follow-ups

- V1: fixed 64-bit backing, implicit `0..N` member values. Explicit widths (`bit_set[Perm; u16]`) and explicit values (`enum E { A = 5 }`) are follow-ups.
- **#1054** (filed): `return match {...}` as a bare expression miscompiles — **pre-existing and independent of enums** (reproduces with a plain `int` scrutinee). The enum test uses the working assign-then-return idiom.

## Testing

- `tests/regression/test_issue1044_enums.ae` — enum values across every position, including the return-index regression.
- `tests/regression/test_issue1046_bit_set.ae` — full operator set.
- Clean build (`make clean` required — `TypeKind` gained `TYPE_ENUM`/`TYPE_BITSET`), C unit suite 231/231, full `.ae` suite 822/825 (the 3 are the known http/h2 network flakes that fail identically on `origin/main` on this box). `HARDEN=1 make ci` compiles clean — no format-security / stack-protector / `-Werror` issues in the compiler C.

Rebased onto `main` after Nic's ranged-match #1047 (v0.366.0) landed; code-independent of it (own AST nodes and `TypeKind`s, cherry-picked with only a trivial CHANGELOG overlap).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the
 next version number before tagging the release.
 
+## [current]
+
+### Added
+
+- **First-class `enum` types and enum-backed `bit_set` masks** (#1044, #1046).
+  Aether gains a real `enum` type — `enum Perm { Read, Write, Exec }` — lowering
+  to a C `int` and usable as a first-class value everywhere: variable
+  declarations (`Perm p = .Write`), function params and return types, struct
+  fields, `==` comparison, and `match` arms (`match p { Write -> ... }`), with
+  implicit selector syntax (`.Read`) wherever the enum type is known. On top of
+  it, `type Mask = bit_set[Perm]` gives an integer-backed set: literals use
+  enum members (`{ .Read, .Write }`) and lower to a `uint64_t` mask; `+`/`-`
+  and `+=`/`-=` are union/difference, `in`/`not_in` test membership,
+  `<`/`<=`/`>`/`>=` are strict/non-strict subset checks, and `card(mask)`
+  lowers to a popcount (a user-defined `card` function of your own is
+  unaffected). Zero runtime cost — an enum is an int, a bit_set is a uint64_t.
+  A `bit_set` over an enum with more than 64 members is a compile error (the
+  fixed 64-bit backing can't hold it); duplicate enum members are rejected.
+  V1 uses a fixed 64-bit backing width and implicit `0..N` member values;
+  explicit smaller widths (`bit_set[Perm; u16]`) and explicit member values
+  (`enum E { A = 5 }`) remain follow-ups. The first natural consumer is
+  capability/flag code (`std.capsicum`'s `R_*`/`F_*` masks) that currently
+  hand-rolls `0x0001`-style const blocks.
+
 ## [0.366.0]
 
 ### Added
--- a/compiler/analysis/typechecker.c
+++ b/compiler/analysis/typechecker.c
@@ -548,6 +548,8 @@ static const char* type_name(Type* t) {
         case TYPE_MESSAGE:  return "message";
         case TYPE_ARRAY:    return "array";
         case TYPE_STRUCT:   return t->struct_name ? t->struct_name : "struct";
+        case TYPE_ENUM:     return t->struct_name ? t->struct_name : "enum";
+        case TYPE_BITSET:   return t->struct_name ? t->struct_name : "bit_set";
         case TYPE_FUNCTION:  return "closure";
         case TYPE_TUPLE:    return "tuple";
         case TYPE_OPTIONAL: return "optional";   // #340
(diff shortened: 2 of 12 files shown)


paul-hammant:
superceded

Issue #1107, opened 2026-07-11
std.http.client: per-request custom CA pin (set_cafile) — verify against a private/self-signed cert, not the system store


paul-hammant:
# Ask: Per-request custom CA in `std.http.client` (pin a private/self-signed cert)

## Request

A way for a `std.http.client` request to **verify the peer against a specific CA/cert
file** instead of the system trust store — i.e. pin a private or self-signed cert
while keeping `SSL_VERIFY_PEER` on. Proposed surface, mirroring `set_insecure`:

```
set_cafile(req: ptr, path: string) -> string   // "" on success, err on null req
```

Semantics: for THIS request, verify the server cert against `path` (a PEM CA/cert
bundle) rather than the default verify paths. Peer + hostname verification stay ON.
Requests that don't set a cafile behave exactly as today (system store).

This is the read-verify (server-cert pin) direction only — NOT client-cert / mTLS.
One per-request setter, mirroring the existing `set_insecure`.

## Motivation

aeo orchestrates a **Proxmox VE host over its HTTPS API (:8006)**, machine-to-machine.
PVE ships a **private CA** (`/etc/pve/pve-root-ca.pem`) — no public chain, and for M2M
we explicitly do NOT want the browser/public-root trust model.

The intended "1970s inter-bank courier" flow (the same doctrine aeo's agent_auth uses):
SSH into the box ONCE, courier PVE's own CA cert back to the orchestrator, then **pin
it** for every subsequent 8006 call. That is STRONGER than what's possible today —
right now the aeo proxmox driver is forced to use `set_insecure(1)` because there is no
way to say "verify, but against THIS cert." We want to drop `set_insecure` and pin.

### What already works (verified)

- `openssl s_client -connect 192.168.0.204:8006 -CAfile pve-root-ca.pem` →
  `Verify return code: 0 (ok)`. The couriered PVE CA verifies the 8006 cert cleanly.
- The 8006 cert's SAN includes `IP Address:192.168.0.204`, so connecting by IP matches
  — no hostname workaround needed once a CA can be pinned.

### The gap (with evidence, `std/net/aether_http.c`)

- The SSL_CTX is process-wide; on Linux it only uses the system store:
  ```
  211: SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, NULL);
  218: if (!load_windows_ca_bundle(ctx)) {
  219:     SSL_CTX_set_default_verify_paths(ctx);
  ```
[...]

Issue #1097, opened 2026-07-10
Selective import std.tcp (subset) suppresses instantiation of omitted tuple wrappers used transitively -> build fails with Undefined function 'tcp_poll2'


paul-hammant:
# Selective `import std.tcp (subset)` in the top-level unit suppresses instantiation of *omitted* wrappers used transitively → build fails with `Undefined function 'tcp_poll2'`

## Summary

When the top-level compilation unit does a **selective import** of `std.tcp` that
omits a tuple-returning wrapper (`poll2` / `read_n` / `write_n`), and that omitted
wrapper is used **transitively** by an imported library module, the wrapper is not
instantiated. Its call site degrades to the raw extern name (`tcp_poll2`,
`tcp_read_n`, `tcp_write_n` — note: **not** the real extern, which is
`tcp_poll2_raw` etc.), and type-checking fails:

```
error[E0301]: Undefined function 'tcp_poll2'
error: cannot destructure — 'tcp_poll2' returns 'UNKNOWN', not a tuple
```

The failure is reported at the **library's** source location, not the top-level
file — which makes it look like the library is broken when it compiles fine on its
own.

Crucially, the presence/shape of the top-level `std.tcp` import is what decides it:

- **No** direct `std.tcp` import in the top-level unit → the transitive wrapper
  instantiates fine → **builds**.
- A **partial** `import std.tcp (connect)` → the omitted wrappers are **not**
  instantiated → **fails**.
- A **full-enough** `import std.tcp (connect, poll2, read_n, write_n)` → **builds**.

So a *selective import that omits a symbol* is stronger than *no import* — it
appears to restrict which `std.tcp` wrappers get code-generated for the whole
program, breaking transitive users of the omitted ones.

## Environment

- `ae 0.379.0 (Aether Language)`
- Linux x86_64

## Minimal reproduction

A library that uses the tuple wrappers internally:

`lib/relaylib/module.ae`
```
import std.tcp (poll2, read_n, write_n)
exports (splicey)
splicey(a: ptr, b: ptr, mx: int) -> string {
    ar, br, e = poll2(a, b, 100)
    if e != "" { return e }
    d, n, r = read_n(a, mx)
    if r != "" { return r }
    w, we = write_n(b, d, n)
    return we
}
```

### P1 — no top-level `std.tcp` import → **BUILDS**
```
import relaylib (splicey)
main() { _ = splicey(null, null, 16); println("ok") }
```
```
[...]

Issue #1092, opened 2026-07-10
std.tcp cannot express a full-duplex relay: no readiness primitive, and read-timeout is indistinguishable from EOF (and fatal)


paul-hammant:
# std.tcp cannot express a full-duplex relay: no readiness primitive, and read-timeout is indistinguishable from EOF (and fatal)

## Summary

`std.tcp` today can only do **half-duplex** stream relaying. Building a correct
bidirectional byte pump (a CONNECT tunnel / TCP splice — client↔upstream, either
side may speak first or stream continuously) is not expressible through the
current stdlib without silent data loss. Two independent defects in
`std/net/aether_net.c`, either one alone sufficient:

### 1. No readiness / multiplexing primitive

`std.tcp` exposes only blocking `read_n`/`write_n` (`recv(2)`/`send(2)`). There is
no `poll`/`select`, no nonblocking mode, and no "read whichever of these fds is
ready." A full-duplex relay must wait on **both** fds at once and service
whichever becomes readable; with only blocking reads from one thread of control
that is impossible — a blocking `read_n(client)` parks while `upstream` has data
waiting, and vice-versa.

Note: the runtime already ships an I/O poller
(`runtime/io/aether_poller_epoll.c`, `aether_poller_kqueue.c`) — it is simply not
surfaced to `std.tcp`. This is an **exposure** gap, not a missing capability.

### 2. Read timeout is indistinguishable from EOF — and is fatal

Every socket is given a fixed 30 s `SO_RCVTIMEO`/`SO_SNDTIMEO`
(`net_set_socket_timeouts`, applied in `tcp_connect_raw` and on accept). When the
receive timeout fires, `recv` returns `-1` with `EAGAIN`/`EWOULDBLOCK`. But
`tcp_receive_n_raw` (and `tcp_receive_raw`) collapse **all** `received <= 0` into
a single branch:

```c
int received = recv(sock->fd, buffer, max_bytes, 0);
if (received <= 0) {
    aether_caps_free(buffer, (size_t)max_bytes);
    sock->connected = 0;                 // <-- marks the socket permanently dead
    return out;                          // <-- returns "connection closed or
                                         //     receive failed" for a TIMEOUT too
}
```

So a **quiet-but-alive** direction (a peer that simply hasn't sent for 30 s — very
normal on a long-lived tunnel) is treated as a closed connection: the socket is
marked `connected = 0` and any subsequent read/write fails. A caller cannot tell
[...]

PR #1113, opened 2026-07-11
docs(changelog): fix stale workflow note + restore [current], drop stray duplicate

paul-hammant:
Pure CHANGELOG hygiene, no content entries changed.

- The top **Workflow** note said "New changes go under `## [0.363.0]`" — stale and self-contradictory (its next line says the pipeline replaces `[current]`). Corrected to `## [current]`.
- No `## [current]` section existed after 0.384.0 released — added the empty section so the next change has a home and the release pipeline has its replace target.
- A duplicate **Workflow** note had been pasted mid-file, buried inside the `## [0.363.0]` section (between its emit_csrc entry and `## [0.362.0]`), left by an old release cycle. Removed.

Reconciled every released version 0.379–0.384 against the merged PRs — all already document their work correctly (#1092, #1097, ae fmt #1100, wrapping64 #1103, set_cafile #1107/#1110). The only genuinely missing entry was the GCC 16 strstr fix, which is handled on its own branch (`fix/gcc16-const-strstr-qualifiers`) where its code lives.

Docs-only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,11 @@ All notable changes to Aether are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-**Workflow**: New changes go under `## [0.363.0]`. When a PR merges to
+**Workflow**: New changes go under `## [current]`. When a PR merges to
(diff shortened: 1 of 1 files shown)


paul-hammant:
Consolidated into #1114 — the CHANGELOG hygiene fixes (stale workflow note, restored [current], removed the stray buried duplicate) now ride along with the GCC 16 strstr fix and its changelog entry, so it's one PR with full CI (this touches .c, so no [skip actions]).

PR #1187, opened 2026-07-18
ci: cache the compiled miniaudio object across matrix runs

paul-hammant:
`std.audio`'s vendored `miniaudio.h` (~96k-line single header) costs **~1 min** to compile via `MINIAUDIO_IMPLEMENTATION`, and `make ci` starts with `clean` — so **every matrix leg paid it on every push** (~9 legs in ci.yml + Windows).

The object is safe to reuse because it depends only on things that don't change per-Aether-commit: the frozen upstream header and the per-runner-fixed compiler. `aether_audio.c` is a thin, stable FFI wrapper with zero Aether coupling.

### Mechanism (surgical — one hot object, not a whole-build cache)

**Makefile:**
- `audio-cache-restore` copies a cached object into `build/` after clean and `touch`es it up-to-date so `make` skips the compile.
- `audio-cache-save` writes it back for `actions/cache` to persist.
- Wired into `ci`: `[0/9]` restore, save after `[1/9]`.
- A content **stamp** (source + `miniaudio.h` + `aether_audio.h` + `$(CC) --version` + `HARDEN` + audio/`EXTRA_CFLAGS` hash) gates reuse. **Mismatch → clean recompile**, so a stale object can never be linked.

**Workflows:** `actions/cache` persists `.ci-cache/` keyed per `matrix.os` + `matrix.cc` + `harden` + `hashFiles(miniaudio.h, aether_audio.c/.h)`. Each OS/compiler permutation (Linux gcc/clang/hardened, macOS ARM64/x86_64, Windows MinGW) gets **its own entry** — objects are never shared across platforms.

### Safety

Two independent layers keep it correct:
1. The `actions/cache` **key** includes OS + compiler → a macOS object is never even downloaded onto a Linux runner.
2. The Makefile **stamp** re-checks compiler identity + build flags before reusing → even a wrong object is rejected → clean recompile.

A key miss (miniaudio bump, or GitHub updating a runner's compiler) rebuilds once, then recaches. **Self-healing, never stale.**

Left `memory-check.yml` alone: it builds with its own `-O0`/ASan `CFLAGS` via `make compiler` (a different object), which would need separate keys for little gain.

### Verified locally
- Restore hit → `make compiler` links with **no** `Compiling std/audio/aether_audio.c` line (compile skipped).
- Corrupted stamp → correctly forces a from-source rebuild (no stale link).

This PR intentionally runs the full matrix (no `[skip actions]`) so the cache populates and a hit can be confirmed on the next push.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,19 @@ jobs:
       if: matrix.install != ''
       run: ${{ matrix.install }}
 
+    # Cache the compiled miniaudio object (std.audio's ~96k-line vendored
+    # header costs ~1 min to compile, and `make ci` starts with a clean).
+    # Keyed per-OS + per-compiler + per-build-variant + the header/source
+    # content, so a hit is only ever a byte-identical object; the Makefile's
+    # audio-cache-restore double-checks a content stamp before reusing it and
+    # falls back to a fresh compile on any mismatch. Never shared across
+    # platforms (matrix.os/matrix.cc are in the key).
+    - name: Cache miniaudio object
+      uses: actions/cache@v4
+      with:
+        path: .ci-cache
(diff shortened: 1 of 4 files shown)


paul-hammant:
Nic to follow up with thoughs after insights sought via email

PR #1201, opened 2026-07-20
ask: emit link requirements from the import graph

paul-hammant:
Design ask from the aether-ui project (precedent: the builder-ctx ask → #725). Three separate undefined-reference hunts in one weekend — std.regex/pcre2, std.http/ssl, contrib.sqlite — all because aetherc resolves the import graph and then drops that knowledge. Minimum ask: a `// aether-link: -l...` header in emitted C (emit-time ground truth, travels with the artifact); optional query twin `ae cflags --libs <src.ae>` on the same resolver. Full writeup + downstream evidence in the added doc.

https://claude.ai/code/session_01Q5pckLB6QCPfjVQAYcE9LZ

### DIFF
--- a/emit-link-requirements-from-import-graph.md
+++ b/emit-link-requirements-from-import-graph.md
@@ -1,0 +1,81 @@
+# Ask: emit link requirements from the import graph
+
+**Status: request + design sketch, from the aether-ui project.**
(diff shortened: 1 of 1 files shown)


paul-hammant:
wrong way to ask, Claude!

Issue #1221, opened 2026-07-21
tools/ae.c (8.3k lines, one TU): split into cohesive sources to speed incremental compile


paul-hammant:
## Problem

`tools/ae.c` is **8,292 lines in a single translation unit** (127 top-level functions). `build/ae` is compiled as one direct `-O2` compile of that file (Makefile:1131), so **any one-character edit to `ae.c` recompiles all 8.3k lines** — the dominant cost in the edit-`ae.c`-and-rebuild loop, which is a hot loop right now (the cross-compile work in #1208/#1216/#1218/#1220 all lives here and gets touched constantly).

Splitting it into a few cohesive `tools/ae_*.c` TUs means an edit recompiles only the touched cluster + relinks. On a warm build that's roughly the difference between a multi-second and a sub-second turnaround.

## Why this is safe (no runtime cost)

- Splitting a TU changes **which file** functions live in, not **what runs**. These are all distinct command-handler functions (`cmd_build`, `run_cross_build`, …), not hot leaf functions.
- `-O2` was **not** inlining across these function boundaries in the single TU either — they're large driver functions.
- The one theoretical loss (cross-TU inlining / IPO) is irrelevant for `ae`: it's a build **driver** that spends ~all its wall-clock in `posix_run` / `zig cc` / disk I/O, not its own code. If ever a concern, LTO on the release `ae` recovers it.

## The pattern already exists

`ae` is **already** a multi-TU link: `tools/ae_help.c`, `tools/ae_fmt.c`, `tools/apkg/toml_parser.c` are separate sources linked into `build/ae` today (Makefile:1131). This is just "add more `ae_*.c` and move functions across."

## Suggested first cut (self-contained, highest-value): `tools/ae_cross.c`

The cross-compile machinery is one cohesive, low-coupling cluster and the code we edit most:
- `cross_target_to_zig` (ae.c:4942), `cross_target_needs_sysroot` (4968), `cross_find_manifest` (4978), `cross_collect_core_list` (5004), `run_cross_build` (5063) — a mostly-contiguous ~4942–5290 span.
- Small shared-state surface (see seams below).

Do this one extraction, measure the warm-rebuild delta on a real single-function edit, and if it's meaningful, continue with the other clusters.

## Natural clusters after that

- `ae_cross.c` — cross-compile (above)
- `ae_pkg.c` — `ae add` / apkg / toml package management
- `ae_cache.c` — `AETHER_CACHE_DIR` override + atomic cache publish
[...]

PR #1255, opened 2026-07-23
feat(ae): bindgen consts, import C macro constants as Aether consts

### LINKED ISSUE #1245: C interop: import simple C macro constants
Aedis currently hand-copies many Redis C macro constants into Aether: flag bits such as SRI_O_DOWN, client flags, log levels, error codes, sizes, and string constants. This is brittle and creates review noise because every copied value has to be checked manually.

Feature request: provide a way to import simple C macros from headers into Aether when they expand to numeric constants, string literals, or other compile-time scalar expressions.

Aedis benefit: fewer duplicated constants in .ae files, easier sync with Redis headers, and lower risk when migrating command bodies that depend heavily on C flag macros.

nicolas-maman:
## Summary

`ae bindgen consts <header.h> [-I dir]... [--match PREFIX] [-o out.ae]`: object-like C macros that expand to scalar constants become Aether `const`s in a generated module. This is the highest effort-to-value item of the C-interop batch from the Aedis port, which hand-copies hundreds of flag macros today and drifts silently when Redis headers change.

Closes #1245

**Stacked on #1253**; once that merges, this PR's diff reduces to the feature itself.

## Design

The C preprocessor does the evaluation, not a regex over `#define` lines:

1. `cc -E -dM` discovers object-like macro names (function-like and reserved-namespace names excluded up front).
2. A probe file run through `cc -E` yields each macro's full expansion with nested macros resolved. The probe keys each line with the macro name as a **string literal**, because a bare occurrence of the name would itself be expanded, which was the first thing that went wrong when tested.
3. A ~150-line evaluator folds integer constant expressions with C precedence (shifts, `|`, `&`, `^`, arithmetic, unary, char literals, `u`/`l` suffixes on 64-bit literals); string literals concatenate; float literals pass through; **everything else is skipped and listed in a comment at the end of the generated file**, visible in review, never silent.

Nothing is executed; both stages are preprocessor-only. On Windows the driver hands bindgen the same WinLibs gcc it builds with. The command lives in its own translation unit per the direction of #1221 rather than growing `ae.c`.

## Proof

From a Redis-shaped fixture:

```
const SRI_MASTER = 1          // (1<<0)
const SRI_FAILOVER = 24       // (SRI_S_DOWN|SRI_O_DOWN), folded through nested macros
const PROTO_MAX_BULK = 536870912   // (512ll*1024*1024)
const GREETING = "hello world"     // "hello" " " "world"
// Skipped (not a scalar constant expression):
//   POINTER_CAST = ((void*)0)
```

and the integration test's ground truth is flag algebra agreeing across the language boundary: `flags.F_A | flags.F_B == flags.F_BOTH` evaluated in Aether against values folded from C.

## Tests and docs

- `bindgen_consts` integration test: 8 imports from a fixture exercising shifts, nested ORs, suffixed 64-bit literals, negatives, string concat, escapes, floats, a function-like macro and a cast; visible-skip assertions; `--match` narrowing; end-to-end import-and-run.
- `docs/bindgen-consts.md`, README link, CHANGELOG, `ae help` line.
- `-Werror` clean on the new TU under both clang and MinGW GCC; `ae.c` stays at zero MinGW strict warnings on this stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ next version number before tagging the release.
 
 ## [current]
 
+### Added
+
+- **`ae bindgen consts <header.h>`, import C macro constants** (#1245).
+  Object-like macros that expand to integer constant expressions, string
+  literals, or float literals become Aether `const`s in a generated module,
+  with `-I` include dirs, `--match PREFIX` narrowing, and `-o` output.
+  The C preprocessor does the evaluation (discovery via `-dM`, full nested
+  expansion via a probe), so flag algebra like `(SRI_S_DOWN|SRI_O_DOWN)`
+  folds to exactly what C sees; nothing is executed. Macros that are not
+  scalar constants are skipped and listed in a comment at the end of the
+  generated file, never silently dropped. Ports that hand-copy C flag
+  constants (the Aedis/Redis case that motivated the issue) can generate
+  them instead.
+
 ### Fixed
 
 - **Format bugs in printf-family extern calls are caught again** (#1252).
--- a/Makefile
+++ b/Makefile
@@ -1128,7 +1128,7 @@ ae: compiler
 	@echo "==================================="
 	@echo "Building ae command-line tool ($(DETECTED_OS)) v$(VERSION)"
 	@echo "==================================="
-	$(CC) -O2 -DAETHER_VERSION=\"$(VERSION)\" -DAETHER_OPENSSL_LIBS='"$(OPENSSL_LDFLAGS)"' -DAETHER_ZLIB_LIBS='"$(ZLIB_LDFLAGS)"' -DAETHER_NGHTTP2_LIBS='"$(NGHTTP2_LDFLAGS)"' -DAETHER_PCRE2_LIBS='"$(PCRE2_LDFLAGS)"' -DAETHER_CASPER_LIBS='"$(CASPER_LDFLAGS)"' -DAETHER_AUDIO_LIBS='"$(AUDIO_LDFLAGS)"' $(if $(AETHER_ENABLE_LLM),-DAETHER_ENABLE_LLM=1) -Itools tools/ae.c tools/ae_help.c tools/ae_fmt.c tools/apkg/toml_parser.c $(if $(AETHER_ENABLE_LLM),tools/llm_shim.c $(LLM_LDFLAGS)) -o build/ae$(EXE_EXT) $(LDFLAGS)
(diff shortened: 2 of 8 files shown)


paul-hammant:
Pushed a fix for the two Windows failures (`f3a18b24`) — both jobs were failing on the same single test, `integration_bindgen_consts_test_bindgen_consts`, with everything else green.

**Root cause: `popen` + `cmd.exe` quoting.** On Windows `popen` runs the command through `cmd.exe /c`, which strips the **first and last quote of the entire line** before parsing. Both bindgen stages build a command starting with a quoted compiler path (correctly, so a path with spaces survives), so cmd.exe rewrote them:

```
'gcc" -E -dM "C:' is not recognized as an internal or external command
```

Discovery returned zero macros → `ae bindgen consts` exited 1 → the module was never written → the test's `import flags` failed with three E0200s. Nothing to do with the preprocessor logic or the evaluator; those work correctly on MinGW.

The fix is a `bg_popen` helper applying the extra outer quote pair cmd.exe documents — it consumes the outer pair and the inner quotes survive. Isolated it with a three-form probe on a real MSYS2/MinGW box:

| form | command | result |
|---|---|---|
| A (current) | `"gcc" -E -dM "C:/…/flags.h"` | **0 lines** |
| B (wrapped) | `""gcc" -E -dM "C:/…/flags.h""` | 425 lines |
| C (bare cc) | `gcc -E -dM "C:/…/flags.h"` | 425 lines |

B over C so paths with spaces still work — which matters here, since on Windows this gets `s_gcc_bin` (the WinLibs gcc), not a bare `cc`. POSIX is untouched (`/bin/sh` has no such rule).

**Why CI only said `(no output)`.** The test ran the bindgen step under `set -e` with stderr going to a log that only the *following* `grep` would have printed — so the non-zero exit killed the script before anything reached the console, and the harness prints nothing for a shell test that dies silently. `ae run`'s stderr also went to `/dev/null`. Both now report before exiting, so the next Windows break of this kind will name itself. Also added `2>NUL`/`2>/dev/null` to the `-dM` stage, which was leaking preprocessor stderr.

**Verified on a real Win11/MSYS2/MinGW box, not just by inference:** reproduced the exact CI failure first, then confirmed `[PASS]` after the fix, correct folding through nested macros (`F_BOTH = 9`), and `ae_bindgen.c` clean under `-Wall -Wextra -Werror` with MinGW GCC. Linux still passes.

One thing I noticed but did **not** change, since it's cosmetic and yours to call: `bg_discover` dumps the baseline with empty include flags but the header with the real ones (`ae_bindgen.c:~340`). Harmless for the subtraction since it only ever *adds* candidates, but if a `-I` path shadows a system header the baseline could miss a predefine and let it through as a const.

PR #1274, opened 2026-07-25
fix(ci): per-test timeout to unstick releases + asn1 general-purpose scrutiny (leak fix)

paul-hammant:
## Summary

Two independent fixes bundled per request — the release-CI one is the urgent half (it unblocks the stuck 0.442.0 release carrying the cross-crypto HMAC fix).

**Goal: builds stop hanging, and a to-main merge issues a clean release again.**

Three commits:
1. `feat` (google-labs-jules[bot], co-authored by @paul-hammant) — extend the ASN.1 module to general-purpose ASN.1.
2. `fix(asn1)` — scrutiny pass: plug a memory leak + refresh docs.
3. `fix(ci)` — per-test timeout in `test-ae` + release-job cap.

## 1. Release hangs (the urgent fix) — `asks/release-hangs-on-flaky-test-no-per-test-timeout.md`

A flaky hanging `.ae` test made `make test-ae` idle until GitHub killed the Linux release build at the ~3h ceiling, so **0.442.0 tagged but never published** — leaving the cross-crypto HMAC fix (#1271) unreleased and a downstream aeo arm64 gate stuck. No single test's runtime was bounded.

- **Per-test timeout** in `run_test.sh`: each compiled `.ae` test wrapped in `timeout ${AE_TEST_TIMEOUT:-120}`, each shell test in `timeout ${AE_SH_TEST_TIMEOUT:-180}`. A hang now **fails fast as `[TIMEOUT] <name>`** (exit 124) and is named in FAILURE DETAILS — so it caps at ~2min instead of ~3h *and* pinpoints the offender on its next trip. Defaults are generous (real EC-curve + http/tcp-server tests run a few seconds; verified no healthy test trips the bound) and env-overridable.
- **Job backstop**: `timeout-minutes: 45` on the release `build` job (healthy ~10min, macOS ARM64 up to ~22), so even a timeout-escaping hang can't idle the job for hours.

Also directly serves the "conserve GH-action minutes" goal: a hung test currently burns ~3h of Linux runner time per failed release; this caps it at minutes.

## 2. ASN.1 general-purpose extension — scrutiny

The extension (high tag numbers, implicit/explicit context tagging, SET canonical sort, ENUMERATED, the string zoo, UTCTime/GeneralizedTime, BER indefinite length) is **byte-exact against Bouncy Castle** — I extracted BC's encoding rules and verified every hard case:

| case | verified vs BC |
|---|---|
| high tag numbers | `9F 22` (34), `BF 87 68` (1000) |
| explicit `[1000]` over INTEGER 1 | `BF 87 68 03 02 01 01` |
| implicit `[34]` | `9F 22 01 01` |
| BMPString / UniversalString | `1E 02 00 41` (UTF-16BE) / `1C 04 00 00 00 41` (UTF-32BE), round-trips |
| ENUMERATED, UTF8, IA5, UTCTime | all match |
| BER indefinite length | parses to inner content |

**One real defect found + fixed:** every string-type reader returned a fresh `""` literal in the `(string, string)` error slot on success, which codegen wraps into a **1-byte heap string the caller discards without releasing** — a 1-byte-per-call leak across all 11 `read_string_type` callers plus `read_bmp_string`/`read_universal_string`. Fixed by returning the already-tracked (empty) `err` variable instead, matching the clean pattern `read_oid` already used. **Verified 0 leaks under valgrind afterward.** Also refreshed the stale "subset RSA/X.509" doc header.

## Verification

- ASN.1: both `test_asn1_general` and `test_asn1_der` pass; multi-type encode+read leak-stress is 0-lost under valgrind.
- Timeout: full `test-ae` passes clean with the default bounds (no spurious `[TIMEOUT]`); a deliberately-hung probe is caught in seconds and named. Makefile recipe parses; build is `-Werror` clean.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -246,6 +246,11 @@ jobs:
         startsWith(github.ref, 'refs/tags/')
       )
     runs-on: ${{ matrix.os }}
+    # A healthy build+test is ~10 min (macOS ARM64 up to ~22). Cap well under
+    # GitHub's multi-hour default so a hung test (see the per-test `timeout`
+    # in Makefile test-ae) can't idle the release job for ~3h before it's
+    # cancelled — the failure mode that left 0.442.0 tagged but unpublished.
+    timeout-minutes: 45
 
     strategy:
       fail-fast: false
--- a/Makefile
+++ b/Makefile
@@ -725,6 +725,12 @@ test-ae: compiler ae stdlib
 	script="$$tmpdir/run_test.sh"; \
 	printf '#!/bin/sh\n'                                                                             > "$$script"; \
 	printf 'f="$$1"; tmpdir="$$2"; root="$$3"\n'                                                    >> "$$script"; \
+	printf '# Portable per-test timeout: GNU coreutils `timeout` on Linux/MSYS2,\n'                 >> "$$script"; \
+	printf '# `gtimeout` on macOS (coreutils via brew); empty when neither exists\n'                >> "$$script"; \
+	printf '# (macOS without coreutils) so the test still runs, just unbounded.\n'                  >> "$$script"; \
+	printf 'if command -v timeout >/dev/null 2>&1; then TO="timeout $${AE_TEST_TIMEOUT:-120}"; \\\n' >> "$$script"; \
+	printf 'elif command -v gtimeout >/dev/null 2>&1; then TO="gtimeout $${AE_TEST_TIMEOUT:-120}"; \\\n' >> "$$script"; \
+	printf 'else TO=""; fi\n'                                                                       >> "$$script"; \
 	printf 'name=$$(echo "$$f" | sed "s|tests/||;s|/|_|g;s|\\.ae$$||")\n'                         >> "$$script"; \
 	printf 'dir=$$(dirname "$$f")\n'                                                                >> "$$script"; \
 	printf 'base=$$(basename "$$f")\n'                                                              >> "$$script"; \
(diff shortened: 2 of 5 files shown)


paul-hammant:
Fixed all three CI failures (caught up to current main first):

**macOS ARM64 + x86_64** — my per-test `timeout` broke macOS, which ships neither `timeout` nor `gtimeout`: every test failed with `timeout: command not found` (exit 127, ~1250 tests). Now the generated `run_test.sh` detects `timeout` → `gtimeout` → unbounded fallback. **Verified on a real macOS box** (confirmed it has neither command; the previously-127 tests now run and pass).

**Windows (MSYS2)** — this was the pre-existing `test_worker` "detached worker drained pending" flake (unrelated to this PR's changes; 904/905 passed). Root cause: the test busy-spun on `worker.pending()` with a fixed iteration cap, and on a contended runner the spinning main thread starved the detached pool thread before it could self-decrement. Fixed by calling `worker.drain(0)` each spin (a scheduling yield via the worker mutex; a safe no-op for detached jobs, which never enter the ready-list). **Hammered 40× on a real Windows/MSYS2 box, zero flakes.**

The macOS breakage was mine and is fixed; the Windows flake was long-standing and is now fixed for good — a bonus win for the 'make CI reliable / conserve minutes' goal this PR is about.

paul-hammant:
Fixed the macOS leaks-gate failure: `test_asn1_general` leaked 4×1-byte blocks (`4 leaks for 64 total leaked bytes`).

**My earlier ASN.1 leak fix was incomplete** — I validated it against scratch stress programs, not `test_asn1_general.ae` itself, so I missed leaks the full test exercises. The macOS `leaks(1)` gate caught them (my valgrind run on the wrong target didn't). Now fixed:
- `read_tlv` (both definite- and indefinite-length success returns), `read_tagged_ext`, `read_enumerated` still returned a bare `""` literal in the `(X,string)` slot → codegen wraps a leaked 1-byte heap string. Returned a tracked empty-string binding / the in-scope `err` var instead.
- Plus a distinct leak in `read_tlv`'s BER indefinite-length loop: the recursive sub-TLV read's empty success string was never released; now `string.release`'d.

**Verified on the actual test binary**: valgrind now 0 lost / 0 errors (was 4); both asn1 tests pass. Also swept the asn1 consumers — `pem`/`asn1_der` are 0-leak; `rsa_pkcs1`'s small leak is its pre-existing OpenSSL-RAND one already budgeted in `leaks_known.txt` (130), unrelated to this change.

PR #1521, opened 2026-08-12
fix(std): report bytes above 0x7F correctly, and align the docs with the code

### LINKED ISSUE #1516: contrib/tinyweb: schema_api fails to build and websocket fails at runtime
**Priority: P2.** Pre-existing on `main`, found while running
`make contrib-check` for PR #1504, and unrelated to that PR.

Two of the seven contrib tests fail on a clean checkout of `main`:

```
FAIL  tinyweb/schema_api     (build)
contrib/tinyweb/module.ae:943:25: error: incompatible integer to pointer
  conversion initializing 'const char *' with an expression of type 'int'
  [-Wint-conversion]

FAIL  tinyweb/websocket      (run, exit 1)
[FAIL] small frame roundtrip (unmask + 2-byte header): got [] want [echo:small]
[FAIL] large frame length: got 0 want 200
```

Reproduced with this branch stashed, so both are on `main` and not caused by any
recent change. Note the second one: WebSocket frame handling is returning empty
payloads and a zero length, which is the same class of runtime rot that #1442
was filed about after the last tinyweb regression.

`make contrib-check` is therefore red on `main` today, which means it cannot
gate anything until these are fixed.

## Acceptance

- Both entries pass.
- Whatever broke the frame roundtrip gets a test that would have caught it.

nicolas-maman:
Two bugs found by the audit, and the audit itself. Everything below was checked
against the code, and every corrected snippet was compiled.

## Bugs

**`string.char_at` reported bytes above 0x7F as negative.** It returned a signed
`char`, so 0x85 came back as -123:

```
before:  char_at(0)=-127  char_at(1)=-123
after:   char_at(0)=129   char_at(1)=133
```

Any binary parser built on it broke on high bytes. `contrib/tinyweb` read a
masked WebSocket header, computed a negative payload length and dropped the
connection instead of echoing the frame, which is how it surfaced. The C
definition also disagreed with the prototype codegen emits
(`codegen.c:4504` declares `int string_char_at(const char*, int)`) and with the
stdlib extern, both of which say `int`; only the implementation said `char`.

**`contrib/tinyweb` did not build at all.** `tw_start` declared no return type
and returned `-1` on one path and `http.server_start`'s string on another, so
the emitted C assigned an int to a `const char*`. Every tinyweb example and test
failed on that one line. It now declares `-> string`, returning `""` on success.
The examples then compared that result against `0`, which asks whether the
pointer is NULL and so never took the success branch; they compare against `""`
now.

Both halves of #1516 are fixed: `tinyweb/schema_api` builds and
`tinyweb/websocket` passes.

## Documentation audit

| Finding | Evidence |
|---|---|
| `stdlib-reference.md` claimed "Complete reference" | 37 of 69 shipped modules appeared in neither it nor `stdlib-api.md` |
| `http.server_listen` documented | No such function; it is `http.server_start` |
| `cryptography.base64_encode` / `base64_decode` documented | Neither exists. Base64 is `std.encoding`, returning `string` / `string!`, not the tuples described. `std.cryptography`'s own header already said "use `encoding.base64_*`" |
| c-interop SQLite extern | Used `callback` as a parameter name, a reserved word, so the declaration could not compile |
| Tutorial 02 | Taught `if (is_prime(num))` on an int-returning function, which the type checker rejects |
| 24 broken relative links | Two `docs/setup/` pages never in the tree, a missing `design/` path segment, a roadmap doc that does not exist |
| README | CI suite is 10 steps, not 9; `ae` list missing `inspect`, `bindgen`, `cflags`, `lib-path` |

The reference now opens with an index of **all 69 modules**, each with its
purpose and export count extracted from the module source, so it cannot drift
out of step with the tree the way the old hand-maintained list did.

The tutorial example was fixed to the idiomatic `-> bool` form and then compiled
and run:

```
2 is prime 3 is prime 5 is prime 7 is prime 11 is prime
```

The changelog archive's dead `contrib/aether_ui/` paths are deliberately left as
written, with a note that the UI library moved to a sibling repo: a changelog is
a record of what happened, not a live index.

## What this pass did not finish

Stated plainly rather than left implied. I compiled all 206 self-contained
`aether` blocks in `docs/` and the README; 57 failed before, 53 after. The
difference is small because most of the remainder are not defects: exercise
stubs with empty bodies, deliberate counter-examples marked `// Wrong`,
fragments that use a name defined in an earlier block, and `@c_import` examples
that need a C source to link. Separating those from real rot needs the blocks to
be marked for what they are, which is a change to how the docs are written, not
a fix. Filed as a follow-up rather than half-done here.

Likewise, 37 modules now have an accurate index entry but not a worked example
section. That is real remaining work and is filed, not hidden.

## Verification

- 230/230 C unit tests, including new coverage asserting `char_at` and
  `char_at_n` return 0..255 for 0x80, 0x81, 0x85, 0xFE and 0xFF.
- `tests/regression/test_char_at_high_bytes.ae` fails against the old
  implementation (`got -128, want 128` and four more) and passes after.
- Full `.ae` suite: 977 passed, 9 failed, all in the known local-environment set.
- 89/89 examples; `fmt_gate` clean.
- `make contrib-check`: `tinyweb/schema_api` and `tinyweb/websocket` both pass.

Closes #1516

### DIFF
--- a/CHANGELOG-archive.md
+++ b/CHANGELOG-archive.md
@@ -1,5 +1,9 @@
 # Changelog — archive (0.259.0 and older)
 
+> Paths under `contrib/aether_ui/` in the entries below no longer resolve.
+> The UI library outgrew this repo and moved to a sibling one; the entries
+> are left as written, since a changelog is a record of what happened.
+
 Older release notes, split out of the active
 [CHANGELOG.md](CHANGELOG.md) to keep the live file small. Format and
 conventions are identical; new entries always go in `CHANGELOG.md`, not
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,58 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the next
 version number before tagging the release.
 
+## [current]
+
+### Fixed
+
+- **`string.char_at` reported bytes above 0x7F as negative** (#1516). It
+  returned a signed `char`, so 0x85 came back as -123 and every binary parser
+  built on it broke on high bytes. `contrib/tinyweb` read a masked WebSocket
+  header, computed a negative payload length, and dropped the connection
+  instead of echoing the frame. The C definition also disagreed with the
+  prototype codegen emits and with the stdlib extern, both of which say `int`;
+  only the implementation said `char`. `char_at` and `char_at_n` now return the
+  byte as 0..255, which is what `bytes.get` has always documented.
+
+- **`contrib/tinyweb` would not build** (#1516). `tw_start` had no declared
+  return type and returned `-1` on one path and `http.server_start`'s string on
+  another, so the emitted C assigned an int to a `const char*` and every
+  tinyweb example and test failed to compile. It now declares `-> string` and
+  returns `""` on success, matching `http.server_start` and the stdlib's error
+  convention. The examples compared that result against `0`, which asked
+  whether the pointer was NULL and so never took the success branch; they now
+  compare against `""`.
+
+### Documentation
+
+- Audited the docs against the code. `docs/stdlib-reference.md` opened with
+  "Complete reference for Aether's standard library modules" while 37 of the 69
+  shipped modules appeared in neither it nor `docs/stdlib-api.md`. It now opens
+  with a generated index covering every module, its purpose and its export
+  count, taken from the source so it cannot drift.
+
+- Corrected API references that no longer matched the code: `http.server_listen`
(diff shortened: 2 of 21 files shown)


nicolas-maman:
Filed the two pieces this pass did not finish, so neither lives only in the PR
description: #1522 (label doc code blocks and gate the complete ones in CI, which
is what would have caught the four rot bugs automatically) and #1523 (worked
examples for the 37 modules that now have an accurate index entry but no
section, plus the stdlib-reference / stdlib-api overlap that let the same base64
error survive in both files).

PR #1529, opened 2026-08-12
test(worker): wait on time, not on a spin count, so Windows CI stops flaking

nicolas-maman:
`test_worker` failed on the **MINGW64** leg with `detached worker drained
pending`, while **UCRT64 passed on the same commit**. Same code, same runner
image, different result: a timing flake, not a defect in the worker pool.

## Cause

All three waits in the test counted iterations for something that depends on
wall-clock scheduling:

```aether
spins = 0
while worker.pending() > 0 && spins < 2000000 {
    worker.drain(0)
    spins = spins + 1
}
```

Each iteration takes the worker mutex, so two million of them elapse in a
fraction of a second. On a two-core runner the cap can expire before the pool
thread is scheduled even once. Worse, the loop is hot: it keeps a core busy
competing with the very thread it is waiting for.

The comment above that loop already described this flake, and an earlier attempt
to mitigate it by using `drain(0)` as a yield point. That was not enough,
because the bound stayed in iterations.

## Fix

One helper, used by all three waits, bounded by time and sleeping between polls
so the pool thread gets the core:

```aether
wait_drained(timeout_ms: int) -> int {
    waited = 0
    while worker.pending() > 0 && waited < timeout_ms {
        _ = worker.drain(0)
        sleep(1)
        waited = waited + 1
    }
    _ = worker.drain(0)
    if worker.pending() == 0 { return 1 }
    return 0
}
```

The assertions after each wait are unchanged, so a pool that genuinely fails to
drain still fails the test; only the patience is now measured in the same units
as the thing being waited for. Normal completion is a millisecond or two.

## Verification

- Six consecutive local runs pass.
- 230/230 C unit tests; `fmt_gate` clean.

Honest limit: the flake only reproduces on a contended Windows runner, so local
runs cannot prove it gone. The argument is structural: a ten-second wall-clock
bound cannot expire before a healthy pool thread is scheduled, and the test no
longer burns a core while waiting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the next
 version number before tagging the release.
 
+## [current]
+
+### Fixed
+
+- **`test_worker` flaked on Windows CI** with "detached worker drained pending".
+  The test waited for the pool thread by counting iterations rather than time,
+  and hot-spun on `worker.drain(0)` while it counted. Two million mutex-takes
+  elapse in a fraction of a second, so on a two-core runner the cap could expire
+  before the pool thread was ever scheduled, and the spinning was itself part of
+  why it was not scheduled. The waits are now bounded by wall-clock time and
+  sleep between polls, so the test stops competing with the thread it is waiting
+  for. The comment above the loop already described this flake; it is now fixed
+  rather than described.
+
 ## [0.522.0]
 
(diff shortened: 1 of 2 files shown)


nicolas-maman:
Closing: this belonged on #1526 rather than in a PR of its own. #1526 was the
red one, and the flake it hit is unrelated to its `ae_cross.c` change, so
splitting the fix into a separate PR left #1526 red for something it did not
cause and could not merge past.

The commit is now on `fix/ae-cross-build-warnings` instead, so that PR goes
green on its own.

PR #1582, opened 2026-08-15
docs(changelog): restore the [current] section the audit merge folded away

nicolas-maman:
The Release workflow has been failing on main since #1568 merged:

```
No '## [current]' section in CHANGELOG.md, so 0.539.0
would ship with nothing recorded.
```

The audit branch carried its entry under `## [current]`. main had meanwhile cut 0.538.0, which renames that heading to the version. The textual merge kept both bodies under the single surviving heading, so the audit entry landed inside an already-tagged section and the file was left with no `[current]` at all.

This moves the audit entry back under `## [current]`. `## [0.538.0]` is now byte-identical to the section tagged as `v0.538.0` (verified by diffing against `git show v0.538.0:CHANGELOG.md`), so nothing that shipped in that release changes.

No code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the next
 version number before tagging the release.
 
-## [0.538.0]
+## [current]
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


paul-hammant:
Caught up with main (per Paul). The conflict resolution surfaced that the fold went deeper than a missing header: the #1568 audit entries — merged *after* v0.538.0 was tagged — were fused under the released `[0.538.0]` section together with the spec/arms entries that genuinely shipped in it. This branch now: restores `[current]` carrying everything awaiting the next release (aeb-asks entries + the audit's Added/Fixed/Removed), and leaves a single `[0.538.0]` with exactly what that release contained. Both branch commits carry bare `[skip actions]` (doc-only) — so expect no checks on this PR head; the diff is CHANGELOG-only.

PR #1596, opened 2026-08-15
feat(windows): zig cross-build knob, AE_TEST_RUNNER hook, fast CI lane

### LINKED ISSUE #1592: WINDOWS=1 zig-cross knob (mirror FREEBSD=1) + AE_TEST_RUNNER hook in ae test/run
First of a two-step plan (with #TBD-workflow) to get a fast Linux-hosted Windows lane in CI. Background: the Windows MSYS2 legs are the slowest jobs in the matrix (~19 min pending while macOS finishes); on a public repo the minutes are free, so the prize is PR turnaround — a ~4-minute Windows compile/test signal that lands before the author context-switches away.

This issue is the CODE half; the workflow half is deliberately mechanical once this lands.

## 1. `WINDOWS=1` cross knob in the Makefile

Mirror the existing `FREEBSD=1` block (Makefile:220-249): build `aetherc.exe` / `ae.exe` / stdlib FOR x86_64-windows-gnu on a Linux host with

```
CC="zig cc -target x86_64-windows-gnu"
```

Differences from the FreeBSD leg that make this one SIMPLER:
- zig bundles the mingw-w64 headers + CRT sources and builds them on demand — **no sysroot fetch step at all** (no aether-crossbuild dependency for this target).

Differences that make it more work:
- Today all Windows handling keys off `findstring MINGW,$(uname)` — i.e. it assumes the build runs ON Windows (MSYS2). The cross path needs the same routing driven by the knob instead: `EXE_EXT=.exe`, `-lws2_32`/`-lbcrypt` etc. on the link line, the Win32 thread-layer selection, `-D__USE_MINGW_ANSI_STDIO=1` (Makefile:284).
- **Capability-lean by design for lane 1**: no OpenSSL/nghttp2 cross (that is the genuinely expensive part and is explicitly out of scope here). The empty-`AETHER_*_LIBS` capability knobs are load-bearing for exactly this, and vendored PCRE2 (`AETHER_VENDOR_PCRE2`) already covers regex without a sysroot. Compiler + pure-computation stdlib is where the test count lives anyway.

Acceptance: on a Linux box with zig, `make compiler ae stdlib WINDOWS=1 ZIG=<path>` produces PE binaries (`file build/ae.exe` says PE32+), and `wine build/ae.exe build hello.ae` works under Wine ≥ 9.

## 2. `AE_TEST_RUNNER` hook in `ae test` / `ae run`

The cargo pattern (`CARGO_TARGET_*_RUNNER=wine64`): an env var (or `--runner=`) that `ae test`/`ae run` prefixes onto the binary it just built instead of exec'ing it directly.

- Empty default → **zero behavior change** for every existing caller.
- Keeps the decision at the edge: no target-awareness threaded through std.spec, no forked `_windows` copies of spec files — the same describe/it blocks run, only the exec strategy differs.
- Probably ~20 lines in tools/ae.c around the existing posix_run/win_run seam.

## Explicitly out of scope

- The workflow itself (companion issue).
- OpenSSL / nghttp2 windows-gnu cross (revisit only if lane 1 proves itself).
- Any change to the MSYS2 jobs — they remain the fidelity tip permanently (Wine is not NTFS, not IOCP, not TransmitFile; and our scar tissue — msvcrt 3-digit exponents, MinGW __thread, Windows HTTP test timing — is exactly the class Wine would miss or fake).

### LINKED ISSUE #1593: windows-cross.yml: fast Linux-hosted Windows lane (zig cross-build + Wine base-of-pyramid tests)
Second half of the plan started in #1592 (which is the prerequisite: `WINDOWS=1` cross knob + `AE_TEST_RUNNER` hook). This half is deliberately mechanical once #1592 lands.

## The job (ubuntu-latest, target ~4-5 min)

```yaml
# sketch
- uses: actions/checkout@v4
- run: sudo apt-get install -y wine        # Wine 10 WoW64: no i386 multilib needed for x86_64-only
- run: ./scripts/get-zig.sh                 # same provisioning + cache pattern as the FreeBSD cross job in ci.yml
- run: make compiler ae stdlib WINDOWS=1 ZIG=$ZIG
- run: file build/ae.exe | grep -q 'PE32+'  # a cross build that silently emitted Linux binaries proves nothing
- run: wineboot --init && wineserver -w     # warm the prefix ONCE, not per-test
- run: AE_TEST_RUNNER=wine make test-ae AE_SWEEP_SCOPE=base   # or equivalent scoping mechanism
```

Reuse the FreeBSD job's cache discipline (restore on PR, save on main-merge miss) for the zig tarball.

## Scope: base of the pyramid ONLY

In: syntax / compiler / regression sweeps + pure-computation stdlib — string, json, math, collections, list, map, intarr, cryptography vectors (wycheproof), zlib. Wine's CRT-level fidelity against mingw is high there, and that is where the test count lives.

**Permanently excluded** (stay MSYS2-only — the fidelity tip):
- `std.fs` / path semantics — Wine's case-insensitivity mapping is not NTFS; exactly the class that goes green under Wine and red on Windows.
- http server / tcp / h2 / WebSocket / sendfile — Windows is IOCP + TransmitFile, our largest behavioural delta, and these are integration-shaped anyway.
- Actor scheduler timing (work stealing, migration convergence, mailbox contention) — timing assertions under an emulation layer buy flake-re-runs, not signal.

The exclusion mechanism should be an explicit allowlist/denylist file next to tests/ae_sweep_prune.txt (visible, reviewable), not scattered per-test guards.

## Why this is worth a 6th platform lane

Not minutes (public repo = free); **ordering**. Today Windows-specific `-Werror`/MinGW-class breakage surfaces ~20 min into the MSYS2 legs, typically after the author has moved on. A ~4-minute lane converts that into an in-review fix. The MSYS2 jobs stay untouched as the source of truth; this lane is allowed to be lower-fidelity because it is only ever an EARLIER signal, never a replacement gate.

### LINKED ISSUE #1594: docs: README's 'enforced down to libc' reads cross-platform; LD_PRELOAD layer is Linux/FreeBSD-only
Small doc-accuracy issue surfaced while scoping the Wine/Windows CI lane (#1592/#1593).

docs/containment-sandbox.md already states the boundary honestly — its platform table says `Windows | Not supported | No LD_PRELOAD equivalent; spawn_sandboxed is a stub`. But README.md line 9 leads with

> enforced from compile time down to libc

and line 15's three-layer story ends at the LD_PRELOAD shim, with no platform qualifier in either place. A Windows reader reasonably concludes all three layers apply to them, when only the compile-time (`--emit=lib` capability-empty) and scope (`hide`/`seal except`) layers do.

Also worth a sentence in docs/containment-sandbox.md while there: the runtime layer can never be a Wine-tested surface either — preloading into the Wine process does not intercept the guest's calls the way the threat model intends, so the LD_PRELOAD layer is Linux/FreeBSD **by construction**, not by backlog. (Relevant now that a Wine test lane is planned: those tests must not be read as exercising containment on Windows.)

Suggested fix: one qualifier in the README ("…down to libc on Linux/FreeBSD; Windows gets the compile-time and scope layers") + the by-construction note in the doc. Nothing else — the doc's platform table is already right.

paul-hammant:
Closes #1592, closes #1593, closes #1594 — the three-part plan for a fast Linux-hosted Windows signal.

**Why**: the MSYS2 legs take ~20 minutes, so Windows-only compile breakage is discovered long after the author context-switched away. This lane compiles the same code paths on a Linux runner in minutes. Minutes are free on a public repo — the prize is **PR turnaround**, and now also **not burning 20 minutes of doomed compute** when the fast lane fails.

## `WINDOWS=1` cross knob (#1592)

Mirrors `FREEBSD=1`, via `zig cc -target x86_64-windows-gnu`. Simpler than the FreeBSD leg: zig bundles the mingw-w64 headers and CRT, so there is **no sysroot to fetch**.

```bash
make compiler ae stdlib WINDOWS=1 ZIG=/path/to/zig
file build/ae.exe   # PE32+ executable (console) x86-64, for MS Windows
```

Capability-lean by design (out of scope: cross-compiling OpenSSL/nghttp2, the genuinely expensive part) — those std features ship as their "unavailable" stubs, exactly as on any box without the libraries. Vendored PCRE2 needs no host library, so `std.regex` survives.

Three host-`uname`-keyed spots had to opt out or the cross link picked up the host's POSIX libs (`zig-lld: unable to find dynamic system library 'pthread'`): `AUDIO_LDFLAGS`, the `-pthread` block, and `__USE_MINGW_ANSI_STDIO` — the last of which turned out to be **correct as-is**, since zig ships mingw-w64's libmingwex.

## `AE_TEST_RUNNER` hook (#1592)

`ae run` / `ae test` prefix the binary they just built with the runner instead of exec'ing it — cargo's `CARGO_TARGET_<triple>_RUNNER` pattern. **Empty by default**, so every existing caller is byte-for-byte unaffected. 45 lines, two call sites.

The point is that it keeps target-awareness at the **edge**: no forked `_windows` spec files, nothing in std.spec or the test sources knows it is under an emulator, and the `AE_SPEC_FORMAT`/`AE_SPEC_REPORT` contract is inherited straight through the wrapper.

## Fast CI lane (#1593)

`windows-cross` lives **inside windows.yml**, not a separate file — `needs:` only works within one workflow, and gating was the point. Both MSYS2 legs and `headers-msvc` now `needs: windows-cross`, so a fast-lane failure skips the slow tier. (`needs:`-skipped jobs report as *skipped*, not failed; the PR still reads red from the fast lane's own failure, so branch protection behaves.)

Steps: provision zig (same cache discipline as ci.yml's FreeBSD leg) → cross-build → **assert the artifacts really are PE** → warm the Wine prefix once → smoke-test hello-world → sweep with `AE_TEST_RUNNER=wine`.

Exclusions are **named in `tests/ae_sweep_prune_wine.txt`**, not skipped silently — fs/path semantics (Wine's case-mapping is not NTFS), sockets/h2/sendfile (IOCP and TransmitFile, not epoll/sendfile), actor-scheduler timing (flake under emulation), and the LD_PRELOAD sandbox. `make test-ae` gained `AE_SWEEP_EXTRA_PRUNE=<file>`, applied to **both** the `.ae` and `.sh` sweeps — the `.sh` sweep had no prune filter at all, so the excluded suites would otherwise have run anyway.

## Build-target stamp (fell out of the work)

One `build/` tree holds one target's objects and archives; mixing ELF and PE dies deep in the link with `unknown file type`, and a cross build silently clobbers `libaether.a` for the next native build. `build/.build-target` now makes the next mismatching build stop immediately with an actionable message. `make clean`/`help` are exempt — the guard must not block its own remedy.

## Docs (#1594)

README's "enforced down to libc" names the platform boundary; `docs/containment-sandbox.md` states Windows is out **by construction** (there is no `LD_PRELOAD`, so this is not a port waiting to be written) and that **a Wine run does not exercise containment at all** — preloading into the wine process intercepts wine's libc calls, not the guest's, so a green run there would be actively misleading. `docs/build-system.md` documents both cross knobs and the runner hook.

## Verification

- Clean `WINDOWS=1` build from scratch → PE32+ `ae.exe` + `aetherc.exe`, correct stamp.
- `AE_TEST_RUNNER` proven to intercept with a logging wrapper on **both** `ae run` and `ae test`; inert when unset or empty.
- Native build unaffected; **390 unit tests pass**; `AE_SWEEP_EXTRA_PRUNE` verified end-to-end (687 → 203 tests with a test list, message reports the count).
- Every pattern in the wine prune list checked to match real paths (no dead entries).
- Full `.ae` sweep: 850 passing. The only failures are `http_server_h2` — **documented as failing on clean main** in `asks/h2-50-stream-stress-framing-layer-error.md` — plus a load-related `http_request_remote_addr` flake that passes in isolation, and `contrib_vulkan_portability` (no GPU on this box).

**Honest gap**: the Wine execution half is unproven locally (no Wine installed here). The cross-build and the runner hook are both proven; CI is the first real exercise of `AE_TEST_RUNNER=wine`. If it needs tuning, expect it in the timeout or the prune list rather than the mechanism.

Incidentally the lane already earned its keep: the cross-build surfaced two Windows-only `-Wsign-compare` diagnostics inside Winsock's own `FD_SET` macro (`SOCKET` is unsigned). They originate in zig's bundled headers rather than our code, so the fast lane deliberately does not promote warnings to errors — that enforcement stays with the MSYS2 legs, against the headers Windows users actually compile with.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,39 +15,40 @@ on:
 #
 # Two tiers (#1592 / #1593):
 #   1. windows-cross — a FAST lane on a Linux runner: `zig cc -target
-#      x86_64-windows-gnu` cross-build (~minutes), then the base of the test
-#      pyramid RUN under Wine. Catches Windows-only compile breakage (missing
-#      Win32 shims, MinGW-class portability bugs) while the author is still
-#      in the PR.
-#   2. build-mingw / headers-msvc — the real thing on windows-latest. These
-#      are the FIDELITY TIP and the actual gate; Wine speaks for neither
-#      NTFS path semantics, IOCP/TransmitFile sockets, nor actor timing.
+#      x86_64-windows-gnu` cross-build of compiler + ae + stdlib, in a few
+#      minutes. Catches Windows-only COMPILE breakage (missing Win32 shims,
+#      MinGW-class portability bugs) while the author is still in the PR.
+#   2. build-mingw / headers-msvc — the real thing on windows-latest, and
+#      the only tier that RUNS anything. The actual gate.
 #
-# The slow tier `needs:` the fast one, so a cross-build or Wine failure skips
-# ~20 minutes of doomed MSYS2 compute. Cost: the MSYS2 legs start a few
-# minutes later on green runs — worth it, since the actionable signal lands
-# at minute ~4 either way.
+# The slow tier `needs:` the fast one, so a cross-build failure skips ~20
+# minutes of doomed MSYS2 compute. Cost: the MSYS2 legs start a few minutes
+# later on green runs — worth it, since a compile break surfaces at minute
+# ~4 instead of ~20.
+#
+# NOT running tests under Wine here is deliberate (#1593 tried it). `ae` is a
+# compile-and-run driver: `ae run` under Wine needs a WINDOWS C toolchain
+# inside the Wine prefix to compile the C it emits — it tried to download
+# 250 MB of MinGW-w64 mid-job. Driving the native `ae` with
+# AE_CC="zig cc -target ..." instead needs a Windows libaether.a living
+# beside the native one, which one build/ tree cannot hold. Executing tests
+# on a Linux runner is therefore a separate piece of work, tracked in its
+# own issue; the compile half stands on its own.
 
 jobs:
   # ============================================================
-  # FAST LANE — cross-build for Windows on Linux, test under Wine
+  # FAST LANE — cross-build for Windows on a Linux runner
   # ============================================================
   windows-cross:
-    name: Windows / cross-build + Wine tests (x86_64)
+    name: Windows / cross-build (x86_64)
     runs-on: ubuntu-latest
     timeout-minutes: 25
 
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install toolchain + Wine
-      run: |
-        sudo apt-get update
-        # Wine 9+/10's WoW64 means an x86_64-only target needs no i386
-        # multilib. Package name varies by image generation.
-        sudo apt-get install -y gcc make bc curl xz-utils wine64 || \
-          sudo apt-get install -y gcc make bc curl xz-utils wine
-        wine --version || wine64 --version
+    - name: Install toolchain
+      run: sudo apt-get update && sudo apt-get install -y gcc make bc curl xz-utils
 
     # Same provisioning + cache discipline as ci.yml's FreeBSD cross leg:
     # restore on every run, save only on a main-merge miss. zig bundles the
@@ -114,39 +115,6 @@ jobs:
         file build/aetherc.exe build/ae.exe
         file build/ae.exe | grep -q 'PE32+' || { echo "ae.exe is not a PE binary"; exit 1; }
         file build/aetherc.exe | grep -q 'PE32+' || { echo "aetherc.exe is not a PE binary"; exit 1; }
-
-    # Warm the prefix ONCE — per-test wineboot costs more than the tests.
-    - name: Warm the Wine prefix
-      run: |
-        export WINEDEBUG=-all
-        wineboot --init
-        wineserver -w
-
-    # Smoke first: if ae.exe cannot build and run hello-world under Wine,
-    # every sweep failure below would be noise from one root cause.
-    - name: Smoke — hello world through ae.exe under Wine
-      run: |
-        export WINEDEBUG=-all
-        printf 'main() { println("hello from cross-built ae") }\n' > /tmp/hello.ae
-        wine build/ae.exe run /tmp/hello.ae 2>&1 | tee /tmp/hello.out
-        grep -q "hello from cross-built ae" /tmp/hello.out
-
-    # The base of the pyramid, under Wine. AE_TEST_RUNNER (#1592) prefixes
-    # each built test binary with `wine`; nothing in std.spec or the test
-    # sources knows it is running under an emulator. The exclusions are
-    # named in tests/ae_sweep_prune_wine.txt, not skipped silently.
-    - name: Run base-of-pyramid .ae tests under Wine
-      env:
-        WINEDEBUG: -all
-        AE_TEST_RUNNER: wine
-        AE_SWEEP_EXTRA_PRUNE: tests/ae_sweep_prune_wine.txt
-        # Wine adds per-process startup latency; the default 120s per-test
-        # budget is tight for the slowest compile-and-run cases.
-        AE_TEST_TIMEOUT: 300
-      run: |
-        make test-ae \
-          WINDOWS=1 \
-          ZIG="${{ steps.prov.outputs.zig }}"
 
     - name: Upload cross-built binaries
       if: always()
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -12,8 +12,152 @@ on:
 # Build the Aether compiler and ae CLI on Windows using MinGW (GCC).
 # MinGW gives us the same GCC/Clang toolchain surface without an MSVC license,
 # while our aether_thread.h Win32 compat layer keeps the runtime fully native.
+#
+# Two tiers (#1592 / #1593):
+#   1. windows-cross — a FAST lane on a Linux runner: `zig cc -target
+#      x86_64-windows-gnu` cross-build (~minutes), then the base of the test
+#      pyramid RUN under Wine. Catches Windows-only compile breakage (missing
+#      Win32 shims, MinGW-class portability bugs) while the author is still
+#      in the PR.
+#   2. build-mingw / headers-msvc — the real thing on windows-latest. These
+#      are the FIDELITY TIP and the actual gate; Wine speaks for neither
+#      NTFS path semantics, IOCP/TransmitFile sockets, nor actor timing.
+#
+# The slow tier `needs:` the fast one, so a cross-build or Wine failure skips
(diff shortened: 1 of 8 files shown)


paul-hammant:
**Scope correction after the first CI run: the Wine execution half is deferred; this PR is now compile-only.**

The run is worth reading as a result rather than a failure — it invalidated a premise in #1593's plan that neither of us had tested:

| step | outcome |
|---|---|
| Install toolchain + Wine | ✅ |
| Provision zig | ✅ |
| Cross-build (`WINDOWS=1`) | ✅ |
| Verify artifacts are PE | ✅ |
| Warm Wine prefix | ✅ |
| **Smoke: hello-world via `ae.exe` under Wine** | ❌ `[ae] GCC not found. Downloading MinGW-w64 GCC (~250 MB)...` |

`ae` is a **compile-and-run driver**. Running `ae.exe` inside a Wine prefix means the *compiler* runs there too, so it wants a **Windows** C toolchain in that prefix to compile the C it emits. Wine alone was never enough.

I checked the obvious alternative before giving up on it: drive the **native** `ae` with `AE_CC="zig cc -target x86_64-windows-gnu"` so only the *resulting binary* goes through Wine. That needs a Windows `libaether.a` beside the native one — confirmed locally, it fails to link against the Linux archive — and one `build/` tree holds exactly one target's archives (the very collision this PR's `.build-target` stamp exists to catch). The sweep step would also have failed independently: `test-ae` shells out to `build/ae`, which is `build/ae.exe` in a `WINDOWS=1` tree.

That is a real piece of design work, not a tweak, so it belongs in its own issue rather than blocking this one.

**What still ships, and why it is still worth it:**
- `WINDOWS=1` cross knob (#1592) — unchanged, proven.
- `AE_TEST_RUNNER` (#1592) — unchanged, proven on both `ae run` and `ae test`. It is exactly what a future execution lane needs; keeping it costs nothing since it is inert unless set.
- `windows-cross` job — cross-builds and asserts the artifacts really are PE. **The gating is intact and was most of the value**: MSYS2 legs `needs:` it, so a Windows compile break skips ~20 min of doomed compute and surfaces at minute ~4.
- `tests/ae_sweep_prune_wine.txt` — kept and reframed as the design record. The durable part was always the *analysis* of which areas a Wine lane must never claim to cover, not the wiring.
- Docs (#1594) — unchanged and, if anything, more accurate now.

CHANGELOG, `docs/build-system.md` and `docs/containment-sandbox.md` all describe what ships rather than what was planned, and the deferred work is written down where the next person will find it.

Happy to close #1593 as partially-delivered and open a follow-up for the execution lane, or leave #1593 open scoped down to just that — your call on which reads better in the tracker.

Issue #1598, opened 2026-08-15
codegen: leading-underscore function passed as a value emits an undeclared name (definition renamed, reference not)


paul-hammant:
## Summary

A top-level function whose name starts with `_`, **passed as a value** (function pointer) rather than called, fails to compile: the definition is renamed `_h` → `ae_h` but the reference site is emitted verbatim, so the generated C references an undeclared symbol.

Found on ae 0.541.0 while upgrading `aeo` to that release. It blocks two of aeo's five binaries (`bin/aeo-agent.ae`, `bin/aeo-supervisord.ae`) — every HTTP route registration is `server_get(raw, "/health", _h_health, 0)`, which is exactly this shape.

## Repro

Three lines, no imports:

```aether
_h_under(req: ptr, res: ptr, ud: ptr) -> int { return 0 }
takes_fp(f: ptr) -> int { return 0 }
main() { takes_fp(_h_under); println("ok") }
```

```
$ ae build m1.ae -o m1
m1.ae:3:28: error: '_h_under' undeclared (first use in this function); did you mean 'ae_h_under'?
    3 | main() { takes_fp(_h_under); println("ok") }
```

The compiler's own "did you mean `ae_h_under`?" is the bug in one line: it renamed the definition and not the reference.

Both conditions are required — either alone is fine:

| case | result |
|---|---|
| `_h_under` **called** (`_h_under(1)`) | builds, runs |
| `h_plain` (no underscore) passed as value | builds, runs |
| `_h_under` **passed as value** | **fails** |

## Cause

`rename_leading_underscore_functions` (compiler/codegen/codegen.c, ~L4045) — the #279/MSVCRT fix that keeps Aether `_write` off the CRT's reserved `_write` — renames the definition and then fixes up references with:

```c
rename_calls_to(program, fn->value, safe);
```

but `rename_calls_to` only rewrites **call** nodes:

```c
if (node->type == AST_FUNCTION_CALL && node->value &&
    strcmp(node->value, from) == 0) { ... }
```

A function used as a value is an `AST_IDENTIFIER` (or whatever the bare-reference node is), not an `AST_FUNCTION_CALL`, so it is skipped and emitted with the original spelling while the definition has moved.

`rename_extern_colliding_functions` (~L4070, the #1366 extern-collision pass) calls the same helper immediately below and looks to have the identical hole — same shape, different trigger. Worth fixing both in one change; I only have a live repro for the underscore path.

## Suggested fix

[...]

PR #1599, opened 2026-08-15
feat: optional why-message on fluent matchers (#1576); ae add from release artifacts (#1360)

### LINKED ISSUE #1360: ae add cannot install from release artifacts, only git clone
`ae add` only ever fetches source over git. There is no way to consume a published release artifact, even though the machinery to *use* one already exists.

Raised by Paul Hammant.

## What `ae add` does today

`tools/ae.c:5555-5557`:

- without a version: `git clone --depth 1 https://<package> <pkg_dir>`
- with a version (`ae add github.com/user/repo@v1.2.0`): a full `git clone`, then `git checkout <tag>`, because the shallow clone will not carry the tag

Packages land in `~/.aether/packages/<package>`. There is no code path anywhere in `ae add` that looks at release assets.

## Why this is a small gap rather than a missing design

The consuming half is already built. The binary-import prepass (`prepare_binary_imports`, `tools/ae.c:2459`) handles `import foo` where no `foo` source module exists but a `libfoo.so` / `foo.so` is on the search path. It reads the artifact's `aether_lib_meta()` catalog, synthesizes an Aether interface stub (`@extern` declarations for the exports, plus `builder` wrappers so trailing-block DSLs keep their call-site shape), and adds the artifact to the link line.

So a released `--emit=lib` artifact is already importable. What is missing is only the fetch.

## Suggested shape

Let `ae add` prefer a release asset when one matches the host triple, and fall back to the git clone otherwise:

```
ae add github.com/user/repo@v1.2.0          # picks the matching release asset if present
ae add github.com/user/repo@v1.2.0 --source # forces the clone
```

Points worth deciding:

- asset naming, presumably the same `<name>-<version>-<platform>` convention the Aether releases already use
- checksum verification, a released binary should be verified in a way a git tag does not need to be
- the binary-import path is POSIX-only today (gated on `dlopen`), so Windows would keep cloning until DLL hosting lands
- source packages stay the default for anything that must be compiled into the same cycle

## Benefit

Faster installs for large packages, no toolchain needed to consume a prebuilt library, and reproducible pinning against an immutable artifact rather than a tag that can move.

### LINKED ISSUE #1576: std.spec: optional "why" message on the fluent value-comparison matchers
Migrated from `aeocha/asks/fluent-matchers-optional-why-message.md` (requested 2026-06-28 by the `aeo` orchestrator's specs). The fluent chain now lives in `std/spec/module.ae` (PR #1574), so the right home for this ask is here, not the aeocha repo.

## Summary

The fluent chain matchers (`to_equal`, `to_equal_str`, `to_be_gt`, `to_be_lt`, `to_contain`, `to_start_with`, `to_be_truthy`, `to_be_falsy`) auto-generate their failure text (`expected '5', got '7'`) and have no way to carry the operator's intent message. The flat asserts already do — `assert_str_eq(a, b, msg)`. Please add an optional `msg` to the fluent matchers so a chain can say *why* a check matters, not just *what* the values were:

```
expect_str(got).to_equal_str(want, "explicit budget wins over within()")
expect_int(n).to_be_gt(0, "a derived attempt budget is never zero")
```

`satisfies(s, pred, msg)` already takes a `msg` — the precedent (and the verdict plumbing) is in the API; this extends it to the value-comparison matchers.

## Why

aeo nearly converted its ~243 spec assertions to the fluent chain and backed off: 158 of them are `assert_str_eq(computed, "expected", "why")` where the "why" is documentation the value can't convey (e.g. `"peer egress -> internal net"`). The current fluent form would render only `expected 'internal', got 'shared'` — losing the thing a failing-CI triager actually needs. So today the choice is fluent chain XOR an intent message, which keeps consumers on the flat API even where a chain would read better.

## Suggested shape

Trailing `msg` on each value-comparison matcher; failure text `"${msg} — expected '${want}', got '${s.value}'"` when `msg != ""`, else unchanged. Aether is fixed-arity (no overloading), so this is a signature change — breaking for existing fluent callers. Two ways to land it:

- **(a)** make `msg` required and sweep the (currently few) fluent callers, or
- **(b)** add parallel `*_because(s, want, msg)` variants and leave the bare ones as-is.

(a) is cleaner while the fluent API is still young; (b) is non-breaking. Your call — you own the API's churn budget.

## Impact

Unblocks aeo (and any consumer whose tests are "assert a computed value equals a constant, and document the rule") from adopting the fluent chain. With it, the chain becomes the strictly-better option and aeo would sweep its ~243 assertions over.

paul-hammant:
Closes #1576, closes #1360 — two independent features in one PR to conserve Actions minutes.

---

## #1576 — optional intent message on the fluent matchers

`to_equal`, `to_be_gt`, `to_be_lt`, `to_be_truthy`, `to_be_falsy`, `to_equal_str`, `to_contain` and `to_start_with` take a trailing `msg` and render `"<msg> — <generated text>"`:

```
FAIL: a node's egress set is exactly its declared whitelist — strings differ at index 14
        expected: "api.github.com, api.anthropic.com"
        actual:   "api.github.com"
                                 ^
```

**The (a)-vs-(b) choice turned out to be a false dilemma.** The issue offered a breaking required-`msg` sweep versus parallel `*_because` variants, and @paul-hammant's comment argued (a) with a strong case (810 live aeo call sites, aeo volunteering to sweep). Both rest on "Aether is fixed-arity (no overloading), so this is a signature change" — but Aether *does* have default arguments, and I verified they resolve correctly through UFCS method-call chaining before relying on it:

```aether
_ = mk(7).to_be_gt(0).to_equal(7, "why this matters").to_be_gt(1, "another")
```

So the parameter is **optional**: no breaking sweep, no doubled matcher surface, no worse-but-shorter default. Existing chains keep byte-identical wording (pinned by a test), and a single chain can annotate only the links that need it. aeo can adopt messages incrementally on its 810 sites rather than in one atomic change — and nothing forces a message onto checks whose values genuinely are self-explanatory.

**The diff-form request is honoured.** Per the comment's nit, `to_equal_str` now reports through the caret-aligned diff (the `assert_str_eq_diff` form) once either string reaches 24 characters — where eyeballing a quoted pair stops working and the index + caret is what actually finds the byte. Short strings keep the compact `expected 'x', got 'y'`. `not_()` negation carries the message too.

## #1360 — `ae add` installs from release artifacts

```bash
ae add github.com/user/repo@v1.2.0            # prefers a matching release asset
ae add github.com/user/repo@v1.2.0 --source   # forces the clone
```

Looks for `<repo>-<tag>-<os>-<arch>.tar.gz` (then `.zip`) under `releases/download/<tag>/` — the layout Aether's own releases use — and **falls back to the git clone when nothing matches**, so a package publishing no artifacts behaves exactly as before.

On the points the issue left to decide:
- **Checksum**: a sibling `<asset>.sha256` is verified when published. A **mismatch is fatal** — nothing installs, exit 1, and deliberately **no fall-back to git**, since silently cloning after a failed verification would defeat the check. No published checksum installs with a warning (matching the git path's own trust level rather than refusing outright).
- **Windows**: the fetch itself is cross-platform (`ae_download`/`ae_extract` already handle both); only *consuming* a binary import stays POSIX-gated, unchanged by this PR.
- **Mirrors**: `AE_RELEASE_BASE_URL` repoints the origin, which is also what makes the path testable without the public internet.

`ae_download`/`ae_extract` are promoted from `ae_version.c` statics to shared helpers rather than duplicated.

Two defects found by *testing* rather than reading, worth flagging since both would have shipped silently:
1. `run_cmd*` exec a **tokenized argv, not a shell**, so my first checksum implementation's `$(...)`/`&&` never executed — every verification "passed" as a mismatch. Rewritten in plain C (also removes the shell-quoting risk).
2. The `https://` origin was hardcoded, making the whole artifact path untestable offline.

## Tests

- **`tests/integration/spec_why_message/`** pins the *failure text*, which a passing suite cannot observe: bare wording unchanged (the regression guard for ~810 downstream message-less call sites), message prefixed, caret column asserted at 34 for a divergence at index 14, and the message surviving `not_()`.
- **`tests/integration/ae_add_release_artifact/`** serves a fake forge over loopback with a redirected `HOME` — hermetic, no network. Five properties: install+verify, fatal mismatch *and no git fall-back*, unsigned-warns-but-installs, `--source` skips, no-asset falls back. **Verified to fail** when the checksum gate is sabotaged.
- `test_spec_module` gains a matcher-coverage `it()` — 11 passing, still valgrind-clean at **0 bytes**.
- `spec_format_reporting`, `std_testing_arms` pass; fmt gate clean.

Docs: `docs/testing.md` gains a "Saying *why* a check matters" section; `docs/getting-started.md` documents the artifact layout, the checksum policy and `--source`.

@paul-hammant — worth relaying to aeo that the sweep they volunteered for isn't needed: they can add messages where they help, at their own pace, and their existing 810 sites keep working untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,40 @@ version number before tagging the release.
 ## [current]
 
 ### Added
+- std.spec's fluent value-comparison matchers take an optional trailing
+  intent message (#1576): `to_equal`, `to_be_gt`, `to_be_lt`,
+  `to_be_truthy`, `to_be_falsy`, `to_equal_str`, `to_contain` and
+  `to_start_with` now accept `msg` and render `"<msg> — <generated
+  text>"` on failure. The generated text says what the values were; the
+  message says why the check matters, which is what a failing-CI
+  triager needs. The issue weighed a breaking required-`msg` sweep
+  against parallel `*_because` variants — neither is necessary: Aether's
+  default arguments resolve through UFCS method-call chaining, so the
+  parameter is optional, existing chains are untouched (their wording is
+  byte-identical), and one chain can annotate only the links that need
+  it. `not_()` negation carries the message too.
+- `to_equal_str` now reports through the caret-aligned diff (the
+  `assert_str_eq_diff` form) once either string reaches 24 characters,
+  where a quoted pair stops being legible and the index + caret is what
+  actually locates the differing byte. Shorter strings keep the compact
+  form.
+- `ae add` installs from a published release artifact instead of always
+  git-cloning (#1360). With `@version` it looks for
+  `<repo>-<tag>-<os>-<arch>.tar.gz` (then `.zip`) under the package's
+  `releases/download/<tag>/`, the layout Aether's own releases use, and
+  falls back to the clone when nothing matches — so a package that
+  publishes no artifacts behaves exactly as before. `--source` forces
+  the clone. A sibling `<asset>.sha256` is verified when published: a
+  MISMATCH is fatal, installs nothing, and deliberately does NOT fall
+  back to git (which would defeat the verification); a missing checksum
+  installs with a warning. `AE_RELEASE_BASE_URL` repoints the origin at
+  an internal mirror. The consuming half already existed — the
+  binary-import prepass reads a `--emit=lib` artifact's catalog and
+  synthesizes its interface — so this closes the fetch gap only.
+
+## [current]
+
+### Added
 - `WINDOWS=1` cross-build knob: build the `ae`/`aetherc` toolchain FOR
   Windows from a Linux host with `zig cc -target x86_64-windows-gnu`
   (#1592). Companion to `FREEBSD=1`, and simpler — zig bundles the
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -167,7 +167,28 @@ ae add github.com/user/repo          # latest from GitHub
 ae add github.com/user/repo@v1.0.0   # specific version
 ae add gitlab.com/user/repo           # GitLab
 ae add codeberg.org/user/repo         # Codeberg
+ae add github.com/user/repo@v1.0.0 --source   # force the git clone
 ```
+
+With `@version`, `ae add` prefers a **published release artifact** when the
+package ships one for your platform, and falls back to cloning otherwise.
+The asset is expected at the same layout Aether's own releases use:
+
+```
+https://<package>/releases/download/<tag>/<repo>-<tag>-<os>-<arch>.tar.gz
+```
+
+(`.zip` is tried too; `<os>-<arch>` is `linux-x86_64`, `macos-arm64`, and
+so on.) A sibling `<asset>.sha256` is verified when present — a
+**mismatch is fatal and installs nothing**, and there is no silent
+fall-back to git, which would defeat the check. An artifact with no
+published checksum installs with a warning.
+
+Prebuilt `--emit=lib` artifacts are directly importable: the binary-import
+prepass reads the artifact's catalog and synthesizes the interface, so a
+released library needs no toolchain to consume. Set
+`AE_RELEASE_BASE_URL` to point at an internal mirror serving the same
+layout.
 
 ## Error Handling
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -127,6 +127,37 @@ Available matchers: `to_equal`, `to_be_gt`, `to_be_lt`, `to_be_truthy`,
 `to_be_falsy` (int); `to_equal_str`, `to_contain`, `to_start_with`
 (string); `satisfies`, `satisfies_str` (both).
 
+#### Saying *why* a check matters
+
+Every value-comparison matcher takes an optional trailing message. The
+generated text tells you *what* the values were; the message tells a
+failing-CI triager *why* the check exists, which is the part the values
+cannot convey:
+
+```aether
+spec.expect_str(egress_fqdn_csv("python_vm"))
+    .to_equal_str(want, "a node's egress set is exactly its declared whitelist")
+spec.expect_int(n).to_be_gt(0, "a derived attempt budget is never zero")
+```
+
+```
+FAIL: a node's egress set is exactly its declared whitelist — strings differ at index 14
+        expected: "api.github.com, api.anthropic.com"
+        actual:   "api.github.com"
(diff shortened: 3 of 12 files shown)


paul-hammant:
Relaying to aeo, and confirming the compat claim empirically rather than taking it on trust — I built this branch and ran aeo's whole suite against it with **zero aeo changes**:

```
files: 48   assertions: 304   failing: 0
```

Those are the ~810 message-less fluent sites the regression guard is for. Backward compatibility holds on a real downstream corpus, not just the pinned wording test.

**You were right and my comment was wrong on the premise.** I argued (a)-vs-(b) while repeating the issue's "Aether is fixed-arity, so this is a signature change" — I took that from the issue text instead of checking, and default args through UFCS chaining is a better answer than either option I put in front of you. Verified it resolves the way the PR says, including partial annotation of a single chain:

```
FAIL: explicit budget wins over within() — expected 9, got 7     # .to_be_gt(0).to_equal(9, "…")
```

**The diff-form nit is honoured exactly as asked** — the long-string case now reports through the caret form *with* the message, which was the specific thing I said we'd get the most out of:

```
FAIL: a node's egress set is exactly its declared whitelist — strings differ at index 14
        expected: "api.github.com, api.anthropic.com"
        actual:   "api.github.com"
                                 ^
```

Short strings keep the compact `expected 'abd', got 'abc'`. I checked the 24-char threshold at the boundary (23 → compact, 24 → caret) since a documented cutoff is worth pinning to observed behaviour.

Worth noting for the changelog: **78 of aeo's 477 `to_equal_str` sites compare strings ≥24 chars**, so they pick up the caret diff *with no edit at all* — the upgrade improves existing failure output before anyone adds a single message. That's a nicer property than the sweep I was volunteering for.

We'll adopt messages incrementally where they carry a rule the values can't (the netpolicy/pf whitelist assertions first). No atomic sweep needed — thanks for turning down the churn.

One small thing while I have you, unrelated to this PR: **#1598** (leading-underscore function passed as a *value* emits an undeclared name — `rename_calls_to` only rewrites `AST_FUNCTION_CALL`, so pointer references dangle). It blocks two of aeo's five binaries on 0.541.0; we're working around it by renaming handlers to the trailing-underscore convention. 3-line repro in the issue, no imports needed. Happy to send a patch if you'd rather not context-switch — the sibling `rename_extern_colliding_functions` looks to share the hole.

Issue #1584, opened 2026-08-15
Proposal: co-located std/<foo>_tests sibling dirs — spec-dogfooded unit tests beside each stdlib module


paul-hammant:
Filed from a design discussion (2026-08-15), on the heels of the aeocha tease-out (`std.spec`, #1574) and the #1568 audit — both of which sharpened the same observation: **the mapping between a stdlib module and its tests is a naming convention only, and it has holes.** `std.http.server.lb` sat at 292 lines with zero tests of any kind until a manual sweep found it; nothing structural makes that visible.

## The proposal

Every `std/foo/` gains a sibling **`std/foo_tests/`** containing spec-based unit-test *programs*:

- `test_*.ae` files with `main()`, written against `import std.spec` (+ the matcher arms where apt) — the dogfooding shape;
- **no `module.ae`** — that is load-bearing: without one the directory is not module-resolvable, so `import std.foo_tests` cannot exist and the stdlib namespace stays clean;
- `tests/integration/` keeps everything cross-module or harness-shaped (shell scripts, server/client fixtures, ports). The sibling dirs are *unit* tests of one module; forcing integration tests in would just clone `tests/` sixty times.

The census rule that pays for it all: **every `std/foo` has a `std/foo_tests` or a one-line waiver.** `ls std | grep -v _tests` against its complement is a complete coverage audit — the `lb` class of gap becomes a one-glance find.

## Mechanics (all small, all known)

1. **Discovery**: `make test-ae` adds `std/*_tests` as a sweep root. `tests/ae_sweep_prune.txt` (fixed-substring matching, post-#1568) needs a care-pass so entries like `/lib/` don't eat test dirs.
2. **Packaging exclusion**: the install ships `std/` wholesale (`share/aether/` source-fallback), so six one-line copy sites exclude `*_tests`: three `cp -r std` arms in `release.yml`, one in `install.sh`, two in the Makefile (`install`, `test-release-archive`). `ae version install` unpacks the release zip → inherits the fix; MANIFEST is C-only → the from-source fallback is untouched.
3. **The guard that makes 2 durable**: one assertion in `test-release-archive` that no `*_tests` directory exists under the unpacked `share/aether/std` — any future copy site that forgets the exclusion turns the smoke test red instead of shipping the test corpus.
4. **Gates already in place**: the fmt gate covers `std/` today; the leak gate keys on test names and is unaffected by location.

## Sequencing

[...]

PR #1608, opened 2026-08-16
test(host/ruby): co-locate an embedding-lifecycle test with the bridge

paul-hammant:
Co-locates a real integration test with `contrib/host/ruby/` rather than putting it under `tests/integration/host_ruby/` — the direction argued in #1584. A change to the bridge and its test is now one diff in one directory.

## What prompted it

You noticed the nightly reports `ae check` of `host/ruby` in **7ms** and asked for the same set to have their tests run. Digging in:

| layer | covers ruby? |
|---|---|
| nightly `contrib_ae_check` sweep | ✅ `ae check module.ae` — type-check only (the 7ms) |
| `make contrib-host-check` [1/2] | ✅ `cc -fsyntax-only` on the C shim |
| `make contrib-host-check` [2/2] demos | ❌ `ruby SKIP (no demo)` |
| `contrib_check.sh` (build **and run**) | ❌ **no `host/*` entries at all** |

**Correction to something I said mid-investigation:** Ruby is *not* uncovered. `tests/integration/namespace_ruby/` exercises the generated FFI SDK over Fiddle (verified passing here), and `tests/sandbox/test_shared_map_all.sh` covers `ruby_run_sandboxed_with_map`. But both drive Ruby from the **outside**. Nothing exercised `ruby_init_host` / `ruby_run` / `ruby_finalize_host` directly — which is exactly how the defect below stayed hidden.

## What the test covers

- evaluation reaches the interpreter;
- **Ruby-side computation reaches stdout** (`(2+3)*7` → `35`) — a bridge that loads libruby but never executes would pass a weaker check;
- a **raised exception propagates as non-zero**, rather than being swallowed (a bridge returning 0 for broken script code is worse than one that cannot run: the caller believes it worked);
- **init is idempotent** while the VM is live.

It performs the bridge's own documented `LIBRUBY_SO` probe (`ruby -rrbconfig -e 'print RbConfig::CONFIG["LIBRUBY_SO"]'`), so it works on Debian-style packaging (unversioned `libruby.so` symlink) and Fedora/Bazzite-style (no symlink) alike, and SKIPs cleanly wherever Ruby is absent.

## What it found

**`ruby_finalize_host()` is terminal, and the bridge doesn't say so.** CRuby's `ruby_finalize()` tears the VM down permanently — `ruby_init()` cannot be called again in the same process. But the bridge's post-finalize `ruby_init_host()` returns **0**, and the next `ruby_run()` segfaults inside libruby:

```
eval:1: [BUG] Segmentation fault at 0x0000000000000000
c:0003 p:---- s:0011 e:000010 CFUNC  :puts
```

Reporting success and then crashing is worse than refusing outright. I've documented the supported lifecycle in `README.md` ("call it once at shutdown, or not at all") and the test **deliberately does not exercise the unsupported path** — it pins the supported one so a future fix has a baseline to change. I did not change the bridge's behaviour here: that's a separate decision (make post-finalize init return non-zero? make finalize a no-op?) and belongs to whoever owns the bridge.

## Wiring

`make contrib-host-check` gains a `[3/3]` phase that **auto-discovers** `contrib/host/*/test_*.sh`, so the next bridge to add a co-located test is picked up with no Makefile change:

```
[3/3] Co-located bridge tests (contrib/host/*/test_*.sh)...
  [PASS] host_ruby: eval, compute, exception-propagates, idempotent init
```

Verified the wiring actually gates: with a deliberately failing test in place, `make contrib-host-check` exits **2** rather than printing and passing.

Since the nightly already runs `make contrib-host-check` as a recorded step, this reaches the cachyos box with no change to the nightly script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the next
 version number before tagging the release.
 
+## [current]
+
+### Added
+- `contrib/host/ruby/test_host_ruby.sh` — the first **co-located** contrib
+  test, sitting next to the bridge it exercises rather than under
+  `tests/integration/` (the direction argued in #1584). Drives
+  `ruby_init_host` / `ruby_run` / `ruby_finalize_host` directly: evaluation,
+  Ruby-side computation reaching stdout, a raised exception propagating as a
+  non-zero return rather than being swallowed, and init idempotence while the
+  VM is live. Performs the bridge's own documented `LIBRUBY_SO` probe so it
+  works on both Debian-style and Fedora-style packagings, and SKIPs cleanly
+  without Ruby.
+  Ruby was not uncovered before this — `namespace_ruby` and the shared-map
+  test both drive it from outside — but nothing exercised the embedding
+  lifecycle directly, which is how the finalize defect below went unnoticed.
+- `make contrib-host-check` gained a third phase that auto-discovers
+  `contrib/host/*/test_*.sh` and runs them, so a bridge that adds a
+  co-located test is picked up with no Makefile change. Verified that a
+  failing co-located test fails the target.
+
+### Fixed
+- Documented that `ruby_finalize_host()` is **terminal**
+  (contrib/host/ruby/README.md). CRuby's `ruby_finalize()` tears the VM down
+  permanently, but the bridge's post-finalize `ruby_init_host()` still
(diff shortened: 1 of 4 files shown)


paul-hammant:
Superseded by #1611, which rewrites this in Aether against `std.spec` per review — see the comparison there. Closing in favour of that.

The substantive change: driving the bridge through `import contrib.host.ruby` rather than a C harness means the test covers the module declarations and contrib-link plumbing, not just the C shim. It also needed #1610's skip verbs to land first, so that a box without Ruby reports *skipped* rather than failed.

Issue #1606, opened 2026-08-16
Module-level function name silently breaks a program that uses the same name as a closure parameter


paul-hammant:
## Summary

Two builds of the same program whose generated C is **byte-identical apart from
one function's name** behave differently: one segfaults at startup, the other
runs. Renaming a module-level function `item` → `zzz_item`, changing nothing
else, is the whole fix.

Found in aether-ui (ae 0.541.0, Linux x86-64, GCC/MinGW-style build via aeb).

## Reproduction

In `ui/module.ae`, a module-level function existed:

```aether
item(_ctx: ptr, label: string, on_press: fn) { … }
```

An app (`examples/table_demo`) uses `item` as a **closure parameter** name, which
is idiomatic here for a table/list row:

```aether
t = table(cols) callback |item: ptr, c: int| {
    p = item as *Person
    …
}
```

With the module-level `item` present the app segfaults at startup. Rename it and
the app runs. Verified in a clean `git worktree`, deterministic across repeated
runs, with a rebuilt binary each time.

## The generated C is the same

```
$ diff <(sed 's/ui_item/ui_zzz_item/g' A_item.c) B_zzz.c | grep -c '^[<>]'
0
```

Zero differing lines after normalising the single identifier. So aetherc's
codegen for everything else — including the closures that use `item` as a
parameter — is unchanged. The closure itself looks correct in both:

```c
static const char* _closure_fn_6(_closure_env_6* _env, void* item, int c) {
    Person* p = ((Person*)(item));
```

## Crash site

```
Thread 1 "table_demo" received signal SIGSEGV
#0  aether_string_data ()
#1  _closure_fn_5 ()          // the table's row-render closure, in ui/module.ae
#2  _closure_fn_1 ()
#3  ui_listbox_update ()
#4  main ()
```

`_closure_fn_5` calls the app's cell callback and passes the returned string to
`ui_text`. It dies in `aether_string_data`, i.e. on a bad string — not on a bad
`item` pointer.

## What I could NOT reproduce standalone

Plain shadowing works. Both of these print the parameter's value correctly, so
the bug is not simply "a global shadows a closure parameter":

```aether
item(label: string) -> int { return 999 }
take(|item: int| { print("item = ${item}") })     // prints 42, correct
```

…including with a `_ctx`-taking global inside a builder block. So the trigger
needs more of the real program than I could reduce to a small file — likely the
combination of a `_ctx`-injected global, a same-named closure parameter, and the
[...]

PR #1658, opened 2026-08-19
fix(codegen): a function parameter must shadow a consuming module's function (#1657)

### LINKED ISSUE #1657: Function parameter silently shadowed by a same-named function in a CONSUMING module — closure and its env discarded
## Summary

A function **parameter** is silently replaced by an unrelated module-level
function of the same name, defined in a **different, consuming module** that
does not export it. The caller's closure — and its captured environment — is
discarded and replaced with a bare-fn adapter. The call compiles, runs,
returns cleanly, and does nothing.

This is a **scoping** bug, not an overload-resolution one. Aether has no
overloading (two same-name functions in a module are a C-level `conflicting
types` error, and docs/testing.md confirms "no receiver-type overloading"), so
there is no signature match to get wrong. There is only a scope lookup, and a
parameter is the innermost binding there is. Nothing outside the function —
least of all a non-exported function in another file — should be reachable
under that name.

Same family as #1606, which fixed the case where a **closure** parameter was
rewritten to a namespaced module function. This is an ordinary **function**
parameter, shadowed from a **different module**, and it survives that fix.

## The evidence

`ui/module.ae` has one definition, unchanged:

```aether
btn(_ctx: ptr, label: string, on_press: fn) {
    boxed = box_closure(on_press)
    handle = aether_ui_button_create(..., boxed)
```

That same source generates two different bodies, depending only on whether the
*consuming app* happens to define a function called `on_press`:

```c
// apps/rubiks_cube — which has a module-level on_press(view, x, y)
void* boxed = _aether_box_closure((_AeClosure){
    .fn = (void(*)(void))_aether_bare_adapter_on_press, .env = NULL });

// examples/roles_demo — which does not
void* boxed = _aether_box_closure(on_press);   // the PARAMETER, correct
```

The app's `on_press` is not exported and is never referenced from `btn`. Its
mere existence in the consuming module rewrites the parameter.

Note the adapter's signature is `(void* _env, View*, double, double)` — the
app's three-argument function — while `btn`'s `on_press` is a zero-argument
callback. So a 3-arg function is installed where a 0-arg one is expected, and
`.env = NULL` discards the captured variables as well.

## Runtime symptom

`apps/rubiks_cube`: clicking Shuffle over the test driver returns
`{"ok":true}`, the GTK handler IS connected (`g_signal_handler_find` returns
non-zero), the handler IS entered, it calls the closure and returns — and the
model never advances:

```
AEUI_CLICK handle=5 type=GtkButton signal_id=51 handler=43
AEUI_HANDLER entered closure=0x556acfada700 fn=0x556a93dca6f0 env=(nil)
AEUI_HANDLER returned
```

`env=(nil)` for a closure that captured two variables is the tell.

Renaming the app's `on_press` → `rc_on_press`, changing nothing else, fixes it
completely: Moves 0 → 20 on the first click.

Under sustained driving the app eventually SIGSEGVs inside
`g_signal_emit_by_name`, jumping to a corrupted callback pointer
(`0x555500000001`) — consistent with a 3-arg adapter invoked through a 0-arg
signature.

## Why this is worth prioritising

The failure is **silent and non-local**. Nothing warns; the generated C is
well-formed; the crash, when it comes, is in GLib rather than anywhere near
the cause. And the trigger is invisible from the call site: an app author
choosing an ordinary name like `on_press`, `on_click` or `label` for their own
helper can break an unrelated library call, with no diagnostic pointing at
either file.

It also means a library's PARAMETER NAMES are effectively part of its public
API surface, which they should not be.

Scope on the affected build: the same binaries pass ~262 other driver
assertions, so ordinary closure dispatch works. Only calls whose parameter
name collides are hit.

## Version window

Reported by the aeci line from a controlled matrix (fresh builds, consistent
`/usr/local`, aeb CAS cleared per step): the only configuration observed to
pass is a ~0.543-era source build; every release at/after the 0.543.0 tag
freezes, on bare metal and in a VM alike. I reproduce it on 0.551.0.

Treat that window as suggestive rather than established — the passing install
was a mid-cycle source build whose exact commit is unrecoverable, and the
mechanism above would bite on any toolchain that lowers the name this way.

## What I could NOT reduce

Two smaller cases do **not** reproduce; the closure runs correctly in both:

* same module: `lib_call(on_press: fn)` plus a module-level `on_press(a, b)`
* two modules: the same split across a library and a consumer

So the trigger needs more of the real shape than I could isolate — candidates
are the `_ctx` builder parameter, the trailing-block/`callback` call form, or
translation-unit size. The generated-C comparison above is the solid evidence,
and it is a one-command check.

## Repro

```
git clone https://github.com/aether-lang-dev/aether-ui && cd aether-ui
aeb apps/rubiks_cube/.build.ae
grep -A1 'int ui_btn' target/build/apps/rubiks_cube/rubiks_cube.c
```

`_aether_bare_adapter_on_press` on that line is the bug; `box_closure(on_press)`
is what `examples/roles_demo` produces from the identical source. No spec, no
display, no GUI interaction needed.

paul-hammant:
Closes #1657.

**Based on `fix/1652-mailbox-alignment-assert` (PR #1656), not `main`** — rebase to `main` once that merges, or merge #1656 first and this retargets cleanly.

A library function's **parameter** was silently replaced by an unrelated module-level function of the same name, defined in a **consuming** module that never exported it. The caller's closure — and its captured environment — was discarded.

The same library source produced two different bodies depending only on what the *app* happened to name its own functions:

```c
// app defines a module-level on_press(view, x, y)
void* boxed = ui_box_closure((_AeClosure){
    .fn = (void(*)(void))_aether_bare_adapter_on_press, .env = NULL });

// app does not
void* boxed = ui_box_closure(on_press);        // the PARAMETER, correct
```

Note the shapes: a **3-argument** function installed where a **0-argument** callback belongs, and `.env = NULL` taking the captures with it. The C is well-formed, the call returns, and nothing happens.

## Cause

After module merging, one program holds every module's functions — including non-exported ones from the consumer. `bare_top_level_fn()` asked *"is there a top-level function with this name?"* to decide whether an argument was a bare function needing an env-ignoring adapter, and **never asked whether the name was a local binding**.

A parameter is the innermost binding there is, so nothing outside the function may be reachable under that name. The lookup now yields to an enclosing parameter or a declared local.

Sibling of #1606, which fixed the closure-parameter case in the module **renamer**. This is the **codegen** half — a separate pass with its own name resolution — and an ordinary function parameter shadowed from a *different* module, which survived that fix.

## The part worth a reviewer's eye

**One of the three adapter call sites carried its own inlined copy of the whole-program scan** instead of calling `bare_top_level_fn`. Fixing the helper left that site substituting exactly as before — my first attempt at this changed nothing, and it took instrumenting all three sites to find out why.

Duplicating a scope lookup is how the rule gets fixed in one place and still violated in another. That site now goes through the helper. If we later want a single shared "is this name locally bound?" predicate spanning both the renamer and codegen, this is the argument for it.

## Reduced further than the issue

The issue reported it could not be reduced below the full aether-ui app, and listed `_ctx`, the trailing-block form, and TU size as candidate ingredients. **None were needed.** It reduces to two files with no GUI and no clone — the missing ingredient is the **cross-module split** (the reporter's same-module attempt correctly does not reproduce).

That reduction is the regression test, pinning three properties in increasing order of legibility:

1. the emitted C passes the **parameter**, not an adapter — the direct check;
2. the library's C is **identical** whether or not the consumer defines a colliding name — the property a library author actually depends on;
3. the program **runs** with its captured variable intact — without which a future regression appears only as a callback that silently does nothing, which is exactly how this hid.

The test asserts on the adapter's *use*, not its presence: the discovery walk still declares an adapter for the consumer's genuine bare function, and that declaration is dead code the C compiler drops. Pinning the declaration would fail on harmless output and teach the next person to weaken the test.

## Testing

- Verified to **fail against the pre-fix compiler**, reproducing the issue's exact line, and pass after
- Passing a **genuine bare function** still works — the fix narrows the rule rather than removing the feature
- #1606's own regression test still passes
- `make test` — 394 passed, 0 failed
- `make test-ae` — 969 passed, 3 failed: the same three an unmodified tree fails (`http_client_forward_proxy`, `http_server_h2`, `windows_crt_symbols`, all environmental)
- `check-docs` clean

## Impact

Until this, a library's **parameter names were effectively part of its public API surface**. An app choosing an ordinary name like `on_press` or `label` for its own helper could break an unrelated library call, with no diagnostic pointing at either file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -158,6 +158,37 @@ version number before tagging the release.
 
 ### Fixed
 
+- **A function parameter was silently replaced by a same-named function in a
+  CONSUMING module** (#1657). After module merging one program holds every
+  module's functions, including non-exported ones from the app. Codegen asked
+  "is there a top-level function with this name?" to decide whether an argument
+  was a bare function needing an env-ignoring adapter, and never checked
+  whether the name was a local binding — so a library's `btn(on_press: fn)` had
+  its PARAMETER replaced by an adapter for the app's unrelated
+  `on_press(view, x, y)`, discarding the caller's closure and its captured
+  environment with `.env = NULL`. The generated C is well-formed, the call
+  returns, and nothing happens; in aether-ui a button connected, fired, never
+  moved the model, and eventually segfaulted inside GLib on a 3-argument
+  adapter reached through a 0-argument signature.
+
+  A parameter is the innermost binding there is, so nothing outside the
+  function may be reachable under that name. The lookup now yields to an
+  enclosing parameter or declared local. Sibling of #1606, which fixed the
+  closure-parameter case in the module RENAMER; this is the CODEGEN half, a
+  separate pass with its own name resolution, and an ordinary function
+  parameter shadowed from a different module.
+
+  Worth recording for whoever touches this next: one of the three call sites
+  carried its own inlined copy of the whole-program scan rather than calling
+  the shared helper, which is exactly how the scope rule could be fixed in one
+  place and still violated in another. That site now goes through the helper.
+
+  Until this, a library's PARAMETER NAMES were effectively part of its public
+  API surface: an app choosing an ordinary name like `on_press` or `label` for
+  its own helper could break an unrelated library call, with no diagnostic
+  pointing at either file.
+
+
 - **`multicore_scheduler.c` now compiles for 32-bit targets** (#1652). The
   `Message` layout assertion was guarded by `#if INTPTR_MAX == INT64_MAX`; the
   `Mailbox` one directly below it was not, so every ILP32 triple failed to
--- a/compiler/codegen/codegen_expr.c
+++ b/compiler/codegen/codegen_expr.c
@@ -55,9 +55,47 @@ static const char* arg_drain_lookup(ASTNode* node) {
 /* A bare reference to a top-level Aether function, as opposed to a closure or
  * an fn-typed variable. Its address is a real C symbol, which is the only thing
  * a C function-pointer field can hold. */
+/* Is `name` a parameter of the function currently being emitted?
+ *
+ * Parameters are AST_PATTERN_VARIABLE children of the AST_FUNCTION_DEFINITION,
+ * ahead of its BLOCK. Closures carry AST_CLOSURE_PARAM instead, so both are
+ * accepted; a closure's own parameters shadow just as a function's do. */
+static int name_is_enclosing_param(CodeGenerator* gen, const char* name) {
+    if (!gen || !gen->current_function || !name) return 0;
+    ASTNode* fn = gen->current_function;
+    for (int i = 0; i < fn->child_count; i++) {
+        ASTNode* c = fn->children[i];
+        if (!c) continue;
+        if (c->type == AST_BLOCK) break;   /* params precede the body */
+        if ((c->type == AST_PATTERN_VARIABLE || c->type == AST_CLOSURE_PARAM) &&
+            c->value && strcmp(c->value, name) == 0) return 1;
+    }
+    return 0;
+}
+
 static ASTNode* bare_top_level_fn(CodeGenerator* gen, ASTNode* node) {
     if (!gen || !gen->program || !node ||
         node->type != AST_IDENTIFIER || !node->value) return NULL;
+    /* #1657: a LOCAL BINDING WINS. This function asks "is there a top-level
+     * function with this name?", and after module merging that program
+     * contains every module's functions — including non-exported ones from a
+     * CONSUMING module. So a library function's parameter `on_press` matched
+     * an app's unrelated module-level `on_press`, and the call site emitted a
+     * bare-fn adapter for the app's function in place of the caller's
+     * closure, discarding .env with it. Silent: the C is well-formed, the
+     * call returns, and nothing happens.
+     *
+     * A parameter is the innermost binding there is, so nothing outside the
+     * function may be reachable under that name. Checking the enclosing
+     * function's own parameters is the scope rule this was missing; the
+     * declared-locals list is consulted too, so a local `on_press = ...`
+     * shadows equally.
+     *
+     * Same family as #1606, which fixed the closure-parameter case in the
+     * module renamer. This is the codegen half, and an ordinary function
+     * parameter shadowed from a DIFFERENT module, which survived that fix. */
+    if (name_is_enclosing_param(gen, node->value) ||
+        is_var_declared(gen, node->value)) return NULL;
     for (int i = 0; i < gen->program->child_count; i++) {
         ASTNode* pc = gen->program->children[i];
         if (pc && (pc->type == AST_FUNCTION_DEFINITION ||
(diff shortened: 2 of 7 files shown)


nicolas-maman:
Independent confirmation from the other side: this bug is what is currently crashing aether-ui's CI, and this branch fixes it.

aether-ui's new pipeline (aether-lang-dev/aether-ui#28) has `apps/rubiks_cube` dying partway through its driver spec on **both** runners, with two different signals for the same fault:

```
macOS  : Bus error: 10
Linux  : Segmentation fault (core dumped)
```

The macOS crash report names the faulting address directly:

```
"signal":"SIGBUS", "subtype":"KERN_PROTECTION_FAILURE at 0x0000000104ff7aec"
x1 = 0x104ff7aec   symbol: _aether_bare_adapter_on_press
atPC = 28 00 00 b9   ->   str w8, [x1]
```

`0x104ff7aec` is inside the binary's own `__TEXT` (image base `0x104FE4000`, size `0x54000`). It is a store through a pointer that is actually the address of `_aether_bare_adapter_on_press`, so the process writes into read-only executable memory and takes the protection fault.

That follows exactly from the substitution described in #1657. The adapter's real signature is

```c
static void _aether_bare_adapter_on_press(void* _env, View* _a0, double _a1, double _a2);
```

three parameters, while a `btn` callback takes none. Every button in the app was handed that adapter in place of its own closure, so clicking one called it under the wrong ABI, `View* _a0` held whatever was in the register, and the first store through it faulted.

Verified before and after by building both compilers from source and compiling the same app:

```
$ wt-main/build/aetherc apps/rubiks_cube/rubiks_cube.ae rk_main.c
$ wt-fix/build/aetherc  apps/rubiks_cube/rubiks_cube.ae rk_fix.c
```

Same source line, `ui/module.ae:874`:

```c
// origin/main
void* boxed = _aether_box_closure((_AeClosure){ .fn = (void(*)(void))_aether_bare_adapter_on_press, .env = NULL });

// this branch
void* boxed = _aether_box_closure(on_press);
```

`_aether_bare_adapter_on_press` goes from 4 occurrences to 2 in the generated C, and the two that disappear are exactly the `_aether_box_closure` call sites. What remains is the adapter's own declaration and definition, which are harmless.

Also compiled `counter`, `calculator`, `form`, `table_demo`, `tumbling_cube`, `frames_demo` and `font_picker` with the patched compiler: 7 of 7 clean, no diagnostics.

**Correction to an earlier version of this comment.** I first wrote that this did not reproduce against the released v0.552.0 and was live only on `main`. That was wrong, and the mistake was mine: the build I checked had gone through `aeb`, which used a different compiler than the one I had switched to. Compiling the app with each release's `aetherc` directly:

```
v0.549.0: bare_adapter_on_press occurrences = 4
v0.552.0: bare_adapter_on_press occurrences = 4
```

So the substitution is present in shipped releases, not new on `main`. `get.sh` installs the latest tag, which is how aether-ui's CI picked it up. That makes this worse than I first described rather than better.

On regression risk for this branch: `make test-ae` on it gives 966 passed, 6 failed. The six are `ci_coverage_smoke`, `namespace_basic`, `namespace_multifile`, `namespace_python`, `namespace_ruby` and `repl`, and a control run of those same six against unmodified `main` fails all six identically here, so they are this sandbox (no gcov, interpreters and the REPL's TTY unavailable) and not this change.

Issue #1653, opened 2026-08-18
std.http.server.lb: no connection reuse in either direction — 3.7x behind nginx with 8% of requests dropped


paul-hammant:
`std.http.server.lb` runs ~3.7x behind nginx and drops ~8% of requests under load. Both symptoms trace to the same thing: **the load balancer closes TCP connections in both directions, once per request.** Two independent defects, each fixable on its own.

The module header already carried this as an unconfirmed suspicion ("SUSPECTED FINDABLE BUG ... most likely NO UPSTREAM KEEPALIVE"). This confirms it in source, adds a second defect on the inbound side that the note did not anticipate, and re-measures on the original hardware.

---

## Measured

Re-run of the module header's 2026-07-12 benchmark on the **same box** (Intel N100, 4 cores) with the **same generator** (`oha 1.15.0`, the binary installed that day). Load gen + LB + backends co-located, cores pinned (gen 0-1, LB+backends 2-3), 150k target rps, 20s, two static nginx backends. All three proxies front the *same* backends.

Aether at `1bb0aff6` (current `main`).

| | rps | success | p50 | p99 |
|---|---:|---:|---:|---:|
| nginx (`keepalive 128`) | 59,955 | 100% | 1.54 ms | 4.11 ms |
| haproxy | 56,179 | 100% | 1.69 ms | 3.78 ms |
| **`std.http.server.lb`** | **16,302** | **92.03%** | 6.66 ms | 10.17 ms |

**The error distribution is the diagnostic**, and it rules out CPU saturation:

```
Error distribution:
  [22412] connection closed before message completed   <-- 86% of all errors
  [3302]  operation was canceled
  [293]   connection error
  [28]    aborted due to deadline

Status code distribution:
  [200] 300109 responses
```

Every request that *completed* returned 200. Nothing fails logically — connections are being torn down mid-response. A CPU-bound proxy queues and slows; it does not close sockets. p50 of 6.66 ms against nginx's 1.54 ms is consistent with paying a TCP handshake per request rather than with being compute-limited.

### On #1517

This run includes Nic's "park idle cores instead of spinning forever" fix, which was the reason for re-measuring. **It did not move throughput.** The gap is 3.7x here versus the ~2.5-2.7x recorded in July. That is expected in hindsight: #1517 targets *"an otherwise-idle program"*, and under a 150k-rps benchmark the scheduler is not idle.

[...]

PR #1684, opened 2026-08-21
perf(http): park idle keep-alive connections in a poller

### LINKED ISSUE #1663: HTTP server: park idle keep-alive connections instead of holding a worker
A worker owns a connection for that connection's whole life, so the number of connections the server can hold open at once is the worker count (`cores * 2`, capped at 64). Keep-alive is worth a lot below that line and catastrophic above it.

Measured on an 8-core box (16 workers), 3000 requests per cell, `ab` against a one-route Aether server:

| concurrent clients | close per response | keep-alive, capacity-aware | keep-alive, unconditional |
|---|---|---|---|
| 4  | 17,800 rps | 50,900 rps | 65,600 rps |
| 8  | 22,100 rps | 80,800 rps | 60,700 rps |
| 20 | 22,500 rps | 59,100 rps | **99 rps** |
| 50 | 22,700 rps | 36,500 rps | (starves) |

The fourth column is what happens with no rule: past the worker count, connections that never reach a worker stall their client. #1653 shipped the third column, which keeps a connection open only while nothing is queued for a worker, so the server degrades to close-per-response exactly when it is saturated instead of starving.

That rule is the right behaviour for this architecture, but it is a ceiling, not a fix. A server holding a few thousand mostly-idle keep-alive clients (a browser fan-out, a mobile fleet, an LB in front of it) spends the whole time above the line, so it never gets the 3x.

## What the fix is

When a response completes and the connection stays alive, hand the fd back to a poller and release the worker. Wake and re-submit when the next request arrives. Concurrency then bounds on file descriptors rather than on threads, which is the difference between 64 connections and 64k.

The pieces that do not exist yet:

- **Per-connection state has to outlive the worker.** `HttpConn` is a stack local in `http_server_drain_connection` today, and it carries the read buffer (with any pipelined bytes already read), the TLS session, and the request counter. It needs to be heap-allocated and owned by the parking table.
- **A poller the parked fds live in.** `AetherIoPoller` already backs the accept loop, and the multi-accept path runs one per core; parked connections would register there (or in a dedicated one) for readability.
- **The idle timeout moves.** It is a per-connection `SO_RCVTIMEO`-shaped wait today; parked connections need a deadline the poller enforces, so a client that goes away is reaped without a thread waiting on it.
- **TLS.** A parked TLS connection may hold buffered plaintext inside the `SSL*`, so "readable" has to consult `SSL_pending` as well as the poller.

## Acceptance

- Keep-alive throughput at 50 and 200 concurrent clients is at least what 8 concurrent clients gets today, rather than falling off past the worker count.
- The capacity rule from #1653 becomes unnecessary and is removed (or reduced to a backstop for the inline no-threads path).
- `benchmarks/http/run_lb_reuse.sh` and the table in `docs/http-server.md` are re-measured.

nicolas-maman:
Closes #1663

## The worker count was the connection count

A worker owns its connection for that connection's whole life, so a client sitting idle between requests held a thread. That put the useful keep-alive ceiling at the worker count: past it, connections that never reached a worker stalled, and #1653 kept the server upright only by refusing keep-alive while another connection was queued, which caps the benefit at the same number by construction.

A connection with no request in flight now goes to a poller thread. It costs a descriptor and a table slot, comes back to a worker when its client speaks again, and is closed when the idle deadline passes in silence.

## Measured

8-core box, 16 workers, 3000 requests per cell, `ab -k`, both servers built from the same tree at the same sitting:

| concurrent clients | worker-held | parked | |
|---|---|---|---|
| 4   | 65,300 rps | 58,400 rps | -11% |
| 8   | 83,300 rps | 78,700 rps | -6% |
| 20  | 59,200 rps | 51,400 rps | -13% |
| 50  | 38,100 rps | **74,400 rps** | **1.95x** |
| 100 | 31,400 rps | **76,500 rps** | **2.44x** |
| 200 | 24,500 rps | **70,000 rps** | **2.86x** |

Zero failed requests at every point.

Below the worker count the handoff costs a few percent, because there was no thread shortage to solve there. That is why the worker waits about two milliseconds for a follow-up request before parking, and only while another worker is free to take other work: a client mid-burst sends again in microseconds, so handing that connection over and back is pure cost, while under saturation the same wait is time another connection spends queued. Four heuristics were measured before this one: parking on the instant gives 29,800 rps at 8 clients, waiting unconditionally gives 8,450 at 200.

Past the worker count the curve stops falling. What used to decay toward the connection-per-request rate now holds near peak.

## A wakeup must not cost the parked count

The first working version found a woken connection by scanning the parked table for its descriptor. That is O(parked) per wakeup, under the lock the poller thread holds, which quietly reintroduces the scaling limit the change exists to remove: at four thousand parked connections and a busy traffic mix it is tens of millions of comparisons a second on one thread.

Lookup now goes through an fd-indexed open-addressed table, so a wakeup costs the same at four thousand parked connections as at four. Deletions leave tombstones and the index rebuilds from the slots once they reach half the capacity, which also guarantees a free entry: left to accumulate, tombstones would eventually leave none, and an insert probe that never terminates is worse than a slow one. The deadline sweep collects expired connections in batches rather than restarting its scan per connection.

The index was fuzzed against a reference model: 4M mixed add/lookup/remove operations over a recycling descriptor space under ASan and UBSan, with periodic full-table invariant checks (every live slot findable, live index entries equal to the table count). Clean.

## What the test caught

`tests/integration/http_park_idle_connections/` opens 200 keep-alive connections and serves two rounds on each. It failed the first time I ran it: round one served 300, round two served **16**, the worker count, because #1653's capacity rule was closing every connection the moment another was queued. The benchmarks up to that point had been measuring close-and-reopen rather than parking. That rule is now gated on there being no parking lot, which is the whole reason it existed.

## Verification

- `make -j8` clean, `make test` 394 passed
- Every `tests/integration/http_*` suite passes on macOS (kqueue): 24 suites including h2, TLS, SSE, reverse proxy, actor dispatch, sendfile
- Rebuilt from clean in a `gcc:13` container and re-run there (epoll): parking, keep-alive, HEAD/middleware and actor dispatch all pass
- Fallbacks exercised: no threads, no poller, or a full lot leaves the connection on its worker, which is the pre-parking behaviour
- `aether_http_parked_connections` added to the metrics output; `docs/http-server.md` and the load balancer's performance note rewritten, the latter having listed parking as the next step

## A bug this PR caused, and how it was found

Both macOS legs failed `http_middleware_d2` and said nothing useful. That test runs under `set -e`, so a curl that exceeds its 5s cap kills the script before any diagnostic and the harness prints `[FAIL] … (no output)`. The stalled request was a static file.

`TCP_CORK` and `TCP_NOPUSH` are not the same contract, and this code treated them as one. Clearing `TCP_CORK` on Linux sends what is queued. Clearing `TCP_NOPUSH` on Darwin and the BSDs does not — the stack sends it when something else prompts output. Something always did: a worker went straight back to `recv()` on the same socket and the read path prompted the send. **A parked connection reads nothing until the client speaks, and the client is waiting for that response**, so the bytes sat in the kernel until the connection closed.

Measured at socket level rather than reasoned about:

| | result |
|---|---|
| keep-alive request for the static file | nothing within 3s |
| same request with `Connection: close` | arrives instantly |
| main, same sendfile path (traced: `sendfile sent=21 want=21`) | arrives instantly |
| cork disabled entirely | arrives instantly |
| `TCP_NODELAY` set after clearing `TCP_NOPUSH` | still stuck |
| cork lifted *before* the body write | arrives instantly |

So on the `TCP_NOPUSH` platforms the cork is lifted before the body write, and the body write pushes headers and body together, which is the coalescing the cork exists for. Linux keeps its current behaviour, where clearing does flush.

### The coalescing still happens

The reorder would be worth little if it cost a segment per response. Counting how often a client gets the whole response in a single read, 200 keep-alive requests for a small file:

| build | one-read responses | reads per response |
|---|---|---|
| uncork before body (this fix) | **200/200** | 1.00 |
| no cork at all (control) | 3/200 | 1.99 |

### The test that should have caught it

None of the sendfile, keep-alive, park or static-range suites did. `http_sendfile` T2 already drives two requests down one connection, but with the 1 MiB blob — larger than the socket buffer, so the kernel pushes it whatever the cork does. The gap was a **small** file, and that is the case added: two requests over one keep-alive connection for a 16-byte file, which cannot complete unless the first response is pushed. On the unfixed build it fails with curl's timeout quoted; on the fixed one it passes. Verified on Linux (`TCP_CORK` path) and macOS (`TCP_NOPUSH` path), and the server still cross-compiles for `x86_64-windows-gnu`.


paul-hammant:
The two macOS failures look reproducible rather than flaky — a retry has already been spent and one leg failed again the same way, so I'd not spend another.

## What CI shows

`macOS / Clang (ARM64)` has now failed **three times across two independent runs** (original + rerun), always the same single test:

```
=== FAILURE DETAILS ===
--- integration_http_middleware_d2_test_http_middleware_d2 (shell test) ---
(no output)

Aether Tests: 980 passed, 1 failed, 981 total
```

`macOS / Clang (x86_64)` failed identically on the first run. Everything else on the matrix is green — Linux GCC/Clang/Hardened, FreeBSD native and cross, Windows, RISC-V, the ASan/LSan and Valgrind legs, the no-pthreads cooperative scheduler.

The `(no output)` is just `set -e` killing the script at the first failed command before it prints anything, so the log won't say more than this.

## Why it doesn't look like the known d1/d2 flake

`http_middleware_d1/d2` are on the load-sensitive list, but that pattern is a *varying* subset failing on a busy box. This is one stable test, on two architectures, surviving a rerun.

Two controls point the same way:

- macOS was green on the three most recent PRs that ran it (#1683, #1678, #1676).
- My own now-closed parking PR (#1689) — same file, same architectural change — passed **both** macOS legs. So macOS + parking isn't broken as such; something is specific to this implementation.

## What I could not determine

I tried to reproduce on a macOS 15.7.7 box and **could not** — but not in a way that clears the branch, because that box turns out to fail `http_middleware_d2` **12 out of 12 runs on pristine `main`** too, with a different symptom (`static_files: /static/hello.txt expected 200; got 404`, against a `/var/folders/...` static root). That looks like a separate traversal-guard problem with macOS's symlinked `/var`, filed separately. It means the box is useless as a control here, so I have no local repro and no mechanism to offer — only the CI evidence above.

I had a kqueue-vs-epoll theory about parked connections not waking; I'm explicitly **not** putting it forward, because the one timeout I saw that suggested it never reproduced, and the rest rested on a misread trace.

## Suggestion

The fastest bisect is to run `tests/integration/http_middleware_d2/` on a macOS box that reproduces it, with parking on and off. This branch has no runtime switch, so that currently means a rebuild each way.

That's the one piece from #1689 that might be worth lifting: an `AETHER_HTTP_PARKING=0` env check next to the default, so both arms run on a single binary. It also turned out to matter for benchmarking — comparing across separately built binaries, the same pre-parking cell measured 80,454 and 72,238 rps on consecutive runs (11% drift) while its neighbour was stable to 0.02%. An early revision of #1689 got tuned against that noise before a single-binary A/B exposed it. Numbers there are from a 4-core box, so they're a different operating point to your 8-core ones, not a contradiction of them.

PR #1689, opened 2026-08-21
HTTP server: park idle keep-alive connections instead of holding a worker

### LINKED ISSUE #1663: HTTP server: park idle keep-alive connections instead of holding a worker

paul-hammant:
Closes #1663.

A worker owned a connection for that connection's whole life, so the number of connections a server could hold open at once was the worker count — `cores * 2`, capped at 64. Keep-alive was worth several times close-per-response below that line and a net loss above it, because connections that never reached a worker stalled behind ones sitting idle in `recv`.

When a response completes and the connection stays alive, the fd now goes to a poller thread and the worker is released; the connection is re-submitted to the pool when its next request arrives. Concurrency bounds on file descriptors rather than on threads.

## Measured

4-core box (8 workers), 6000 requests per cell, `benchmarks/http/run_keepalive_parking.sh`:

| concurrent clients | park=off | park=on | delta |
|---|---|---|---|
| 4  | 70,087 rps | 69,804 rps | −0% |
| 8  | 66,247 rps | 61,521 rps | −7% |
| 16 | 34,327 rps | **76,771 rps** | **+124%** |
| 50 | 31,025 rps | **84,523 rps** | **+172%** |

Roughly 7% for the handoff below the worker count, against 2–2.7x above it, and the cliff at `cores * 2` is gone.

The benchmark toggles parking on one binary rather than comparing against a separately built pre-parking one, because the reference number moves under you that way: the same pre-parking cell measured 80,454 and 72,238 rps on consecutive runs while its neighbour was stable to 0.02%. An early version of this PR was tuned against that noise, which is what the middle commit corrects.

## The pieces

- **`HttpConn` moves to the heap.** It was a drain-loop stack local, and it must outlive the worker — it carries the read buffer with any pipelined bytes, the TLS session, and the request counter (which was itself a stack variable).
- **A park poller thread.** One per process. The poller reports only the fd — the epoll backend discards its `actor` cookie ("actor mapping is handled by the caller") — so parking keeps an fd-indexed table for O(1) resume rather than scanning.
- **The idle timeout moves.** `SO_RCVTIMEO` only bounds a thread already sitting in `recv`, so parked connections get a deadline the poller sweeps on its 100 ms tick.
- **TLS.** A connection is parked only when neither the read buffer nor `SSL_pending` holds bytes; already-decrypted plaintext would never make the fd readable.
- **Parking engages only once workers are scarce** — anything queued, or three quarters busy. Parking unconditionally cost 30% at 4 and 8 clients, where nothing is contended and the handoff buys nothing.
- **The #1653 capacity rule stays as a backstop** for builds without parking (no threads, or Windows), where the original starvation is still reachable. With parking it would stop parking from ever engaging under the load it exists for.

## Tests

`tests/integration/http_keepalive_parking/` holds 3x the worker count of simultaneous keep-alive connections and asserts both success and reuse. Success alone proves nothing — without parking the server closes between requests, curl silently reconnects, and all 288 requests still return 200. The TCP-connect count is what discriminates: **72 connects parked** (one per client, identical across five runs) against **174 unparked**.

Verified on both boxes: 72 connections / 24 workers on a 12-core dev box, 24 / 8 on the 4-core CachyOS box.

The rest of the HTTP suite is green (45 tests). Two failures are pre-existing and reproduce with `AETHER_HTTP_PARKING=0`: `http_server_h2` (the known h2 framing-layer error, `asks/h2-50-stream-stress-framing-layer-error.md`) and `http_client_forward_proxy` (port contention).

## Not covered

Parking is POSIX-only — Windows keeps the capacity rule. The `AetherIoPoller` `actor` cookie stays unused rather than widening a runtime interface for this. Parked connections use a dedicated poller rather than sharing the per-core accept pollers; that sharing is a follow-up once the semantics have proven out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the next
 version number before tagging the release.
 
+## [current]
+
+### Changed
+
+- **HTTP server: idle keep-alive connections no longer hold a worker**
+  (#1663). A worker owned a connection for that connection's whole life, so
+  the number of connections a server could hold open at once was the worker
+  count (`cores * 2`, capped at 64) — keep-alive was worth several times
+  close-per-response below that line and a net loss above it, because
+  connections that never reached a worker stalled behind ones sitting idle in
+  `recv`. When a response completes and the connection stays alive, the fd now
+  goes to a poller thread and the worker is released, so concurrency bounds on
+  file descriptors rather than on threads. Measured on a 4-core box (8
+  workers), 6000 requests per cell: 33,513 → 66,671 rps at 16 concurrent
+  clients and 30,890 → 74,474 at 50, against roughly 7% for the handoff below
+  the worker count. Parking engages only once workers are scarce, is POSIX-only
+  (Windows and no-threads builds keep the previous capacity rule), and
+  `AETHER_HTTP_PARKING=0` turns it off without a rebuild.
+
 ## [0.563.0]
 
 ### Fixed
--- a/benchmarks/http/run_keepalive_parking.sh
+++ b/benchmarks/http/run_keepalive_parking.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+# What connection parking is worth to the HTTP server (#1663).
+#
+# Runs the same load against the same binary twice — parking on, then off
+# via AETHER_HTTP_PARKING=0 — so the comparison isolates "parking engaged"
+# from "parking code present" and from build-to-build drift. Comparing
+# against a separately built pre-parking binary does not: the reference
+# number moves under you (one cell varied 11% run to run that way).
+#
+# Parking releases a worker across a connection's idle gap, so the effect
+# only appears above the worker count (cores*2, min 8, max 64). Expect a
+# few percent of handoff cost below that line and a large gain above it.
+#
+# Needs oha (https://github.com/hatoo/oha); wrk or ab would do as well,
+# and the ratio between the two columns matters more than the absolutes.
+set -u
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+ROOT="${ROOT:-$(cd "$SCRIPT_DIR/../.." && pwd)}"; cd "$ROOT" || exit 1
+OHA="${OHA:-$(command -v oha || echo "$HOME/.cargo/bin/oha")}"
+REQ="${REQUESTS:-6000}"
+CONC="${CONCURRENCIES:-4 8 16 50}"; PORT="${PORT:-18310}"
+
+[ -x "$OHA" ] || { echo "ERROR: oha is not installed (set OHA=/path/to/oha)." >&2; exit 1; }
+[ -x build/ae ] || { echo "ERROR: build/ae is missing (run make)." >&2; exit 1; }
+
+# Parking bounds concurrency on file descriptors, so the default soft
+# limit becomes the ceiling it is meant to remove.
+ulimit -n 65536 2>/dev/null || ulimit -n 8192 2>/dev/null
+TMP=$(mktemp -d); SRV=""
+cleanup(){ [ -n "$SRV" ] && kill -9 "$SRV" 2>/dev/null; rm -rf "$TMP"; }
+trap cleanup EXIT
+
+# One route, fixed small body — the same backend the lb reuse benchmark
+# uses, with nothing else on the path.
+./build/ae build benchmarks/http/lb_reuse_backend.ae -o "$TMP/b" >/dev/null 2>&1 || {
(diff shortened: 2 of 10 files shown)


paul-hammant:
Closing in favour of #1684, which got here first (opened ~20 hours before this branch started) and is the more thorough implementation. I didn't check origin for existing work on #1663 before starting — my miss.

Worth recording that the two arrived at the same design independently, including the same two mistakes:

- **Both shipped an O(n) scan of the parked table first and had to replace it.** #1684: "the first working version found a woken connection by scanning the parked table for its descriptor." Same here, replaced with an fd-indexed table. #1684's open-addressed version with tombstones and rebuild — fuzzed against a reference model, 4M ops under ASan/UBSan — is the better answer, and handles descriptor recycling and probe termination properly.
- **Both found #1653's capacity rule hides parking completely.** #1684's test "served 16, the worker count" until the rule was gated on there being no parking lot. Identical finding here, identical fix. Anyone benchmarking parking without gating that rule is measuring close-and-reopen.

Cost below the worker count also agrees closely: −6 to −13% there, −0 to −7% here.

Two pieces from this branch might be worth folding into #1684 as a follow-up:

**1. `AETHER_HTTP_PARKING=0` — a runtime off-switch.** It makes a parking-on/parking-off A/B measurable on a *single binary*, which turned out to matter: comparing across separately built binaries, the same pre-parking cell measured 80,454 and 72,238 rps on consecutive runs (11% drift) while its neighbour was stable to 0.02%. An earlier revision of this branch was tuned against that noise before the single-binary A/B exposed it. It also gives a deployment a way to back the behaviour out.

**2. Asserting on TCP-connect count rather than requests served.** With parking off, the server closes between requests, curl transparently reconnects, and every request still returns 200 — so a success-count assertion passes either way. Connect count discriminates directly: 72 connects parked (one per client, identical across five runs) against 174 unparked, for 72 clients × 4 requests. #1684's round-two starvation check does discriminate, this is just a more direct signal.

Measurements from this branch, for whatever cross-validation is worth — 4-core box, 8 workers, 6000 requests per cell, parking toggled on one binary:

| concurrent clients | park=off | park=on | delta |
|---|---|---|---|
| 4  | 70,087 rps | 69,804 rps | −0% |
| 8  | 66,247 rps | 61,521 rps | −7% |
| 16 | 34,327 rps | 76,771 rps | +124% |
| 50 | 31,025 rps | 84,523 rps | +172% |

Independent hardware (4 cores rather than 8), and the cliff lands at exactly `cores × 2` on both — same boundary #1664 found, so it's structural rather than a tuning artifact.

PR #1691, opened 2026-08-22
test(std): spec-based unit tests for eight untested modules (#1584)

paul-hammant:
Another tranche of #1584, following batches 1–3 (#1674 and its predecessors).

## What this adds

Eight `std` modules had **no test of any kind** — the `std.http.server.lb` class of gap #1584 exists to surface. 53 cases across them:

| module | cases | covers |
|---|---|---|
| `uuid` | 9 | v4/v7 canonical shape, version nibble, RFC 9562 variant, sortability |
| `ksuid` | 4 | 27-char base62 shape, alphabet, second-resolution ordering |
| `ulid` | 5 | 26-char Crockford base32, the excluded I/L/O/U, ordering |
| `tsid` | 5 | 13-char shape, alphabet, high-nibble epoch headroom |
| `nanoid` | 7 | default and explicit lengths, URL-safe alphabet, rejected lengths |
| `actors` | 8 | registry bind/lookup/rebind/unregister/clear contracts |
| `lzf` | 9 | round-trip fidelity, compressibility refusal, negative-length guards |
| `zlib` | 6 | deflate/inflate at every level, gzip framing, magic bytes |

Modules with zero coverage anywhere go from **eight to two**. The remaining pair want a harness rather than a unit test: `casper` is FreeBSD-only Capsicum, `log` writes to stdout.

## Why these are a bit different from batches 1–3

All 36 existing co-located tests are files *moved* from the central suites, and none import `std.spec`. These are the first written **against `std.spec`**, which is the dogfooding half of the proposal that had not started.

They also exercise the failure-path matchers from #1678 — `assert_ok`, `assert_err`, `assert_err_contains`. Every module here returns `(value, err)`, which is precisely the convention those matchers were added for, so this doubles as their first real use.

## Two behaviours worth knowing

Both had to be measured rather than read off the source, and are now pinned:

- **`lzf` refuses any input it cannot shrink**, whatever its length. It first looks like an 8-byte floor — an all-`x` string compresses at 8 bytes — but `"abcdefgh"` is refused, and so is a 24-byte all-distinct string. A caller feeding it random or already-compressed bytes hits this routinely and gets only `"lzf compress failed"`.
- **`zlib.gzip_inflate` rejects a raw deflate stream** rather than returning garbage, so the two framings are not silently interchangeable.

## Verification

Every suite was checked against a **mutation of the module it covers** — a version nibble, a returned length, an alphabet, a dropped guard — to confirm it fails when the module is wrong rather than merely passing today. One early mutation (`bytes.new(27)` → `26` in `ksuid`) produced no failure; that turned out to be an inert scratch buffer rather than a weak test, and mutating the actually-returned `bytes.finish(rev, 27)` does fail the length assertion.

`ae fmt --check` clean on all eight, all pass locally, and none are caught by `tests/ae_sweep_prune.txt`.

## No infrastructure changes

`find std -name 'test_*.ae'` is already a sweep root, and `make install` already deletes `std/**/test_*.ae` — both landed with earlier batches, so this tranche is test files only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 `main`, the release pipeline automatically replaces `[current]` with the next
 version number before tagging the release.
 
+## [current]
+
+### Added
+
+- **Spec-based unit tests for eight previously untested `std` modules**
+  (#1584). `uuid`, `ksuid`, `ulid`, `tsid`, `nanoid`, `actors`, `lzf` and
+  `zlib` had no test of any kind — the `std.http.server.lb` class of gap the
+  issue was filed about. 53 cases, co-located as `std/<module>/test_*.ae`,
+  and the first co-located tests written against `std.spec` rather than
+  relocated from the central suites, which is the dogfooding half of the
+  proposal. Modules with zero coverage anywhere drop from eight to two
+  (`casper` is FreeBSD-only, `log` writes to stdout). Two behaviours had to
+  be measured rather than read off the source and are now pinned: `lzf`
+  refuses any input it cannot shrink, whatever its length, reporting only
+  `"lzf compress failed"`; and `zlib.gzip_inflate` rejects a raw deflate
+  stream rather than returning garbage.
+
 ## [0.563.0]
 
 ### Fixed
--- a/std/actors/test_actors_registry.ae
+++ b/std/actors/test_actors_registry.ae
@@ -1,0 +1,99 @@
+// std.actors unit tests (#1584) — co-located with the module.
+//
+// The module had no test of any kind before this. The registry is a
+// name -> actor_ref map with a small set of documented edge cases:
+// rebinding a name overwrites without error, unregister reports
+// whether it removed anything, and whereis on a missing name yields
+// null. Those contracts are what this covers.
+//
+// registry_clear exists for exactly this purpose, so each `it` starts
+// from an empty registry via before_each rather than depending on the
+// order the cases run in.
+
+import std.spec
+import std.actors
+
+message Ping {}
+
+actor Probe {
+    state n = 0
+    receive { Ping() -> { n = n + 1 } }
+}
+
+main() {
+    fw = spec.init()
+
+    spec.describe(fw, "std.actors registry") {
+        spec.before_each() callback {
+            actors.registry_clear()
+        }
+
+        spec.it("starts empty after a clear") callback {
+            spec.assert_eq(actors.registry_size(), 0, "size is 0")
+        }
+
+        spec.it("binds a name and looks it up") callback {
+            a = spawn(Probe())
+            actors.register("probe", a)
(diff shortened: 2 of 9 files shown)


paul-hammant:
Correction: the claim in this PR's description and commit message that these eight modules "had **no test of any kind**" is **wrong**. Every one of them was already covered.

My survey globbed `tests/{regression,unit,integration}/*<module-name>*` by filename only, and never looked inside any file. That misses a test whose filename does not contain the module's name — which is exactly how all eight were covered:

| module | pre-existing test | why my glob missed it |
|---|---|---|
| `uuid`, `ksuid`, `ulid`, `tsid`, `nanoid` | `tests/regression/test_ids.ae` (110 lines) | filename says `ids`, not `uuid`/`ulid`/… |
| `actors` | `tests/regression/test_std_actor_registry.ae` (168 lines) | filename says `actor`, my glob wanted `actors` |
| `lzf` | `tests/integration/lzf_roundtrip/probe.ae` (124 lines) | the *directory* is named, the file is `probe.ae` |
| `zlib` | `tests/integration/zlib_roundtrip/probe.ae` (174 lines), `test_zlib_gzip.ae` (50 lines) | same |

626 lines of existing coverage, not zero. `grep -rl "import std.<mod>" tests/` finds all of it in one command, which is what I should have run.

Worse than the miscount, the overlap is substantial:

- **`actors`**: the existing test covers a **superset** of my suite — `register`, `unregister`, `is_registered`, `registry_size`, `registry_clear` **and `whereis`**, which mine does not touch at all.
- **`uuid`**: `test_ids.ae` already asserts the same three things my v4 cases do — 36-char length, version nibble `'4'` at position 14, variant nibble in `{8,9,a,b}` — just in the older exit-on-failure style rather than through `std.spec`.

So the accurate description of what landed here is: **spec-style suites for modules whose existing coverage was in the central directories under names that did not match the module**, with genuinely new coverage in places (the `lzf` compressibility refusal, `zlib`'s gzip-vs-raw framing rejection, nanoid's `generate_n` rejection paths, ULID's excluded `I/L/O/U`) and duplicated coverage in others.

That does not make the tests worthless — the co-location and the `std.spec` dogfooding were the point of #1584, and the discoverability problem this exposes is *itself* the thing #1584 was filed about: `std/actors` ↔ `tests/regression/test_std_actor_registry.ae` is a convention that a mechanical audit cannot follow, which is precisely the argument for co-locating. But the justification I gave was false, and "eight modules with zero tests" should not stand in the changelog or the history.

Two follow-ups I'd suggest, happy to do either:

1. **Fix the changelog entry** on `main` — it currently repeats the false claim.
2. **Retire the duplicated central tests** rather than leaving both. `test_ids.ae` and `test_std_actor_registry.ae` now overlap heavily with the co-located suites; the honest resolution is to port `whereis` coverage into `std/actors/test_actors_registry.ae` and delete the originals, which is the migration #1584 actually asks for — instead of the third test home it warns about.

My census script for future tranches greps imports rather than filenames.

PR #1693, opened 2026-08-22
test(std): retire duplicated central tests into the co-located suites (#1584)

paul-hammant:
**This PR now repairs #1691 as well as adding to it.** Original scope was `snapshot`/`list`/`map`; it now also corrects a false claim I made in #1691 and completes the migration that claim was hiding.

## The mistake

Both #1691 and this branch's first commit said the modules they covered "had no test of any kind". That was **wrong for every one of them**.

My census globbed `tests/{regression,unit,integration}/*<module>*` by filename and never read a file, so it missed coverage filed under a name that doesn't match the module. `grep -rl "import std.<mod>" tests/` finds it all in one command:

| module | existing test | why the glob missed it |
|---|---|---|
| `uuid`,`ksuid`,`ulid`,`tsid`,`nanoid` | `tests/regression/test_ids.ae` (110 ln) | filename says `ids` |
| `actors` | `test_std_actor_registry.ae` (168 ln) | says `actor`, glob wanted `actors` |
| `lzf` | `integration/lzf_roundtrip/probe.ae` (124 ln) | dir is named; file is `probe.ae` |
| `zlib` | `zlib_roundtrip/probe.ae` (174 ln), `test_zlib_gzip.ae` (50 ln) | same |
| `snapshot` | `integration/snapshot_cow/probe.ae` (246 ln) | same |

Ironically this *is* the problem #1584 exists to solve: `std/actors` <-> `tests/regression/test_std_actor_registry.ae` is a mapping no mechanical audit can follow.

## What this PR does

Rather than leave two test homes per module — the outcome #1584 explicitly warns against — every central assertion is ported into the co-located suite and the original deleted.

### Before -> after mapping

| retired | replaced by | result |
|---|---|---|
| `tests/regression/test_ids.ae` | `std/{uuid,ksuid,ulid,tsid,nanoid}/test_*.ae` | 30 cases |
| `tests/regression/test_std_actor_registry.ae` | `std/actors/test_actors_registry.ae` | 11 cases |
| `tests/integration/lzf_roundtrip/` | `std/lzf/test_lzf_roundtrip.ae` | 11 cases |
| `tests/integration/test_zlib_gzip.ae` | `std/zlib/test_zlib_roundtrip.ae` | 10 cases |
| `std/snapshot/test_snapshot_cell.ae` **(deleted)** | `snapshot_cow/probe.ae` (kept) | — |

Each of the 8 cases in `test_std_actor_registry.ae` is annotated in the new suite with the case it ports (`// --- ported: case 4 (...)`) so the mapping is checkable in the diff.

### Kept deliberately

- **`tests/integration/zlib_roundtrip/`** — drives a C shim and `fs.read_binary` to exercise the AetherString unwrap path inside `aether_zlib.c`. That's an integration concern a unit suite can't reach, and the reason the directory has a `shim.c`.
- **`tests/integration/snapshot_cow/`** — 246 lines covering all five snapshot functions in more depth than my suite did. **My `snapshot` suite is deleted instead**, being pure duplication.

## Coverage gained

The ported cases had to be at least as strong as what they replaced, and in one place they're considerably stronger:

**`actors`** — the regression test's case 2 ("looked-up ref is usable for `!` send") and case 8 (cross-context routing) printed PASS **without asserting anything**. Its own comment: *"No direct way to peek at the listener's state from outside"*. A send that silently went nowhere would have passed.

Both now assert against a **spy actor** that records what it was told. Mocking is the idiomatic answer here — and asserting on the spy rather than on the subject also exercises the reply path (`from ! Tell`) that reading the subject's own state would skip.

Verified the assertions bite: stubbing `whereis` to return null fails **4 cases including those two**, where the original would still have printed PASS.

Also ported, closing gaps my suites had:
- **`lzf`** — binary round trip over `0x00..0xFF`, and a buffer with embedded NULs. Text round trips prove little about a byte codec.
- **`zlib`** — empty-input round trip, empty-buffer rejection, garbage rejection, corrupt-gzip rejection.

## Verification

- **78 co-located cases** pass across the 10 suites
- **`tests/regression`: 306/306** pass after the deletions
- Both kept integration tests pass (`zlib_roundtrip: 6 cases`, `snapshot_cow`)
- `ae fmt --check` clean on all of `std/*/test_*.ae`
- No dangling references to the removed paths in `Makefile`, `ae_sweep_prune.txt`, or CI config
- The `list`/`map` suites still catch mutations (`get` always null -> 3 assertions fail each)

## Changelog

The `[current]` entry is rewritten — it repeated the false "no test of any kind" claim — and gains a `### Changed` entry describing the retirement.


paul-hammant:
Same defect as #1691, which I've corrected there — flagging it here **before** this merges rather than after.

My survey globbed `tests/{regression,unit,integration}/*<module-name>*` by filename and never looked inside a file. `grep -rl "import std.<mod>" tests/` gives the real picture:

| module | pre-existing | overlap with this PR |
|---|---|---|
| `snapshot` | `tests/integration/snapshot_cow/probe.ae` — **246 lines** | **near-total** |
| `list` | 52 files import it | mostly incidental use, no dedicated suite |
| `map` | 29 files import it | mostly incidental use, no dedicated suite |

The `snapshot` case is the bad one. That probe's own header enumerates eight cases:

```
//   2. store publishes the new value AND returns the displaced old one.
//   4. cas success swaps when `expected` matches, returns the displaced
//   7. null-cell defence on load / store / cas.
//   8. new(null) is legal (empty cell) and free is null-safe.
```

That is **exactly** what my nine cases assert, and its 246 lines cover the same five functions in more depth than my 136. The "no test of any kind" claim in the commit message is false for `snapshot`, and the suite is largely redundant.

`list` and `map` are the defensible part: the 52 and 29 importing files are overwhelmingly *using* the containers to build something else, not testing them, and I found no dedicated suite for either. The out-of-range/unbound-key wart those tests pin does look genuinely unasserted anywhere.

**How I'd like to proceed, unless you'd rather otherwise:**

1. Drop `std/snapshot/test_snapshot_cell.ae` from this PR — `snapshot_cow/probe.ae` already does the job, and a second copy is the "three test homes" problem #1584 explicitly warns about. If co-location is wanted for `snapshot`, the honest move is to *move* the existing probe, not write a parallel one.
2. Keep the `list` and `map` suites, and fix the commit message so it stops claiming an absence of coverage I did not verify.

Happy to push that as an amend to this branch. Tell me if you'd rather keep all three and just fix the wording.

For future tranches the census greps imports, not filenames.

PR #1733, opened 2026-08-24
asks: inline the std.mem scalar accessors; add an offset bulk copy

paul-hammant:
Spec-only feature request from the ebiten-in-aether port (Apache-2.0 downstream, so no code is reproduced — measurements and API surface only).

**Problem:** `std.mem` scalar accessors lower to extern calls into `libaether.a`; a per-pixel compositing loop (~30–40 accessor touches/pixel) costs ~460ns/pixel — 141ms for a 300-sprite frame vs 0.6ms for the same loop in C. The call boundary is the cost, and no `-O` level can inline across the static-library seam.

**Asks:**
1. Codegen-lower `get/set_byte`, `_int`, `_long`, `_float32/64` (+ `_sz` twins) to direct loads/stores, preserving the documented null-`ptr` contract inline; keep the extern symbols for ABI/FFI.
2. `mem.copy_at(dst, dst_off, src, src_off, n)` — today interior-span bulk copies are impossible (no pointer arithmetic, and `mem.copy` starts at offset 0), so scanline copies are byte loops.
3. (Optional) an `_unchecked` escape hatch — probably unnecessary if 1's inline branch is enough; measure first.

**Acceptance:** self-contained micro-benchmark in the ask (two 640×480×4 buffers, read-combine-write per pixel, 60 frames): ~8–9s today, target within ~2–3× of C at `-O2`.

With 1+2, the downstream engine deletes its only C helper file and returns to pure Aether behind an already-pinned test suite.

https://claude.ai/code/session_01C6HGvRyHHXspoj6RBT7Hh9

### DIFF
--- a/asks/inline-mem-accessors-for-pixel-loops.md
+++ b/asks/inline-mem-accessors-for-pixel-loops.md
@@ -56,10 +56,11 @@ loads/stores vectorizes and/or reduces to `memcpy`-class throughput.
    opaque blit) are memcpy-shaped, and byte-at-a-time is the only current
    spelling. Same null/permissive contract as `mem.copy`.
 
-3. *(Optional, only if 1 wants a stricter contract)* An unchecked-access
-   escape hatch in the spirit of `floatarr_get_unchecked` for loops that
-   have hoisted their own null/bounds proof. With ask 1's inline null
-   branch we likely don't need it — measure first.
+3. *(Downgraded after measurement — see Acceptance)* An unchecked-access
+   escape hatch in the spirit of `floatarr_get_unchecked`. Revisit only if
+   a profile of ask 1's result shows the inline null branch is the
+   remaining cost; the review benchmark (inlined-with-null-check 38ms vs
+   raw-C 6ms per 60 frames, both far inside frame budget) says it isn't.
 
 ## Call-site census (downstream, by family)
 
@@ -74,11 +75,33 @@ loads/stores vectorizes and/or reduces to `memcpy`-class throughput.
 ## Acceptance
 
 A self-contained micro-benchmark, no downstream code needed: allocate two
-640×480×4 byte buffers; per "frame", for every pixel read 4 bytes from
-each buffer, combine, write 4 back; run 60 frames. Today that's ~8–9s.
-Target: within ~2–3× of the same loop hand-written in C at `-O2`
-(i.e. tens of milliseconds, not seconds). `tests/` presumably wants that
-loop as a perf regression probe alongside the correctness ones.
+640×480×4 byte buffers (via `std.arena` or a caller-supplied buffer —
+`std.mem` itself has no allocator); per "frame", for every pixel read 4
+bytes from each buffer, combine, write 4 back; run 60 frames.
+
+Measured baseline for exactly that loop (Linux x86-64, gcc 12.2,
+ae 0.578.0 — from the PR review, confirmed by objdump showing 8 `call
+aether_mem_get_byte` instructions in the inner loop):
+
+| | 60 frames |
+|---|---:|
+| Aether today (`std.mem` accessors) | ~215 ms |
+| C `-O2`, inlined, null check kept | ~38 ms |
+| C `-O2`, raw indexing | ~6 ms |
+
+**Target: within ~6× of raw C — i.e. the cost of retaining the null
+contract inline, ≈40 ms per 60 frames on a modern x86-64.** That is
+checkable, and honest about what inlining can and cannot recover; the
+null branch itself is the residual and is explicitly acceptable.
+
+(The 141ms-per-frame figure in the Symptom table is a *different, heavier*
+body — the downstream compositor touches ~30–40 accessors per pixel
+including float64 scratch and blend math, where this acceptance loop
+touches 12. Both shrink by the same mechanism; the acceptance bar is the
+minimal loop above, with the measured baseline.)
+
+`tests/` presumably wants that loop as a perf regression probe alongside
+the correctness ones.
 
 With asks 1+2 landed, the downstream engine folds its C helper back into
 pure Aether behind an already-pinned test suite — the "runtime stays C"
--- a/asks/inline-mem-accessors-for-pixel-loops.md
+++ b/asks/inline-mem-accessors-for-pixel-loops.md
@@ -1,0 +1,86 @@
+# Codegen-inline the `std.mem` scalar accessors (and give `mem` an offset copy)
+
+**From:** the ebiten-in-aether port (2026-08-24) · **Where it bit:** the
+software 2D compositor at the heart of a game engine — per-pixel RGBA work
+over a 640×480 framebuffer at 60Hz.
+
+**Note on this spec:** the downstream repo is Apache-2.0 (this one is MIT),
+so this ask is specification and measurements only — no downstream code is
+reproduced here. Everything below is describable from `std.mem`'s own
+public surface.
+
+## Symptom
+
+A pure-Aether inner loop that composites sprites (read 4 source bytes,
+read 4 destination bytes, blend in float, write 4 bytes — roughly 30–40
+`std.mem` accessor touches per pixel) costs **~460ns per pixel**. Concrete
+frames from the engine's benchmark, same machine, same gcc:
+
+| Workload (one frame) | Pure Aether | Same loop as a C helper |
+|---|---|---|
+| 300 sprites of 32×32, source-over | 141 ms | 0.6 ms |
+| One rotated 320×240 blit, bilinear | 27 ms | 3.8 ms |
+
+A 60Hz game has a 16ms budget for *everything*, so the pure-Aether form is
+~9× over budget before game logic runs. The downstream port had to keep one
+C file containing only these inner loops; every other line of the engine is
+Aether. We'd like to delete that file.
+
+## Why it's slow (and why `-O2` doesn't save it)
+
+`mem.get_byte` / `mem.set_byte` / `mem.get_float64` etc. each lower to a
+call into `libaether.a`. The work inside is a couple of instructions; the
+call is the cost. At ~10M+ accessor touches per frame this is tens of
+millions of cross-TU function calls per second, and the C compiler cannot
+inline them across the static-library boundary at any `-O` level (short of
+LTO, which the toolchain doesn't do today). The same loop with direct
+loads/stores vectorizes and/or reduces to `memcpy`-class throughput.
+
+## Asks
+
+1. **Lower the scalar `std.mem` accessors in codegen** instead of emitting
+   extern calls: `get_byte`/`set_byte`, `get_int`/`set_int`,
+   `get_long`/`set_long`, `get_float32`/`set_float32`,
+   `get_float64`/`set_float64` (and the `_sz` twins). Semantics must be
+   preserved exactly — including the documented null-`ptr` behaviour
+   (read returns 0 / write returns 0) — an inline null test is a branch,
+   not a call, and predicts perfectly in a loop. The extern symbols must
+   remain in the runtime for existing binaries and FFI consumers; this is
+   purely a call-site lowering.
+
+2. **`mem.copy_at(dst, dst_off, src, src_off, n)`** (name yours). Aether
+   deliberately has no pointer arithmetic, which means today there is no
+   way to bulk-copy between *interior* spans of two buffers —
+   `mem.copy(dst, src, n)` can only start at offset 0. Row-wise image
+   operations (copy a scanline of a sub-region, the fast path of an
+   opaque blit) are memcpy-shaped, and byte-at-a-time is the only current
+   spelling. Same null/permissive contract as `mem.copy`.
+
+3. *(Optional, only if 1 wants a stricter contract)* An unchecked-access
+   escape hatch in the spirit of `floatarr_get_unchecked` for loops that
+   have hoisted their own null/bounds proof. With ask 1's inline null
+   branch we likely don't need it — measure first.
+
+## Call-site census (downstream, by family)
+
+- Per-pixel composite loops (byte accessors, ~30–40 per pixel): sprite
+  blitting, rect fills, triangle rasterization with per-vertex color.
+- PNG scanline unfiltering (byte accessors, sequential).
+- Bitmap-font glyph blits (byte accessors).
+- Interpolation scratch (float64 accessors, 4–16 per pixel).
+- Small state tables — input snapshots, board games (byte/int accessors;
+  not hot, but they'd ride along).
+
+## Acceptance
+
+A self-contained micro-benchmark, no downstream code needed: allocate two
+640×480×4 byte buffers; per "frame", for every pixel read 4 bytes from
+each buffer, combine, write 4 back; run 60 frames. Today that's ~8–9s.
+Target: within ~2–3× of the same loop hand-written in C at `-O2`
+(i.e. tens of milliseconds, not seconds). `tests/` presumably wants that
+loop as a perf regression probe alongside the correctness ones.
+
+With asks 1+2 landed, the downstream engine folds its C helper back into
+pure Aether behind an already-pinned test suite — the "runtime stays C"
+line can retreat to where it belongs (devices, GUI toolkits, syscalls),
+rather than covering arithmetic on bytes.


paul-hammant:
I reproduced the ask's acceptance benchmark on this box (Linux x86-64, gcc 12.2, ae 0.578.0) to check the claim before it goes any further. **The core case holds and the diagnosis is right — but one number in the ask is pessimistic, and the correction makes the ask easier to accept, not harder.**

## Measured

The ask's own acceptance shape: two 640×480×4 buffers, per frame read 4 bytes from each, combine, write 4 back, 60 frames.

| | 60 frames | per frame |
|---|---:|---:|
| **Aether today** (`std.mem` accessors) | **215 ms** | 3.6 ms |
| C at `-O2`, direct indexing | **6 ms** | 0.1 ms |
| **C at `-O2`, inlined but keeping the null check** | **38 ms** | 0.6 ms |

So the gap is **~36×**, not the ~1400× the ask's "today that's ~8–9s" implies. I suspect that figure came from a debug build or a heavier per-pixel body than the acceptance loop describes; either way the acceptance criterion should be restated against a measured baseline, or the first implementer will think they have already succeeded.

**The 36× is still the whole argument, and it survives the correction.**

## The diagnosis is confirmed at machine-code level, not inferred

`objdump` on the benchmark binary shows **8 actual `call` instructions to `aether_mem_get_byte`** in the inner loop. Every accessor in `std/mem/module.ae` is a one-line wrapper over an extern:

```aether
get_byte(p: ptr, i: int) -> int { return aether_mem_get_byte(p, i) }
```

so there is nothing for gcc to inline across the `libaether.a` boundary. That is exactly what the ask says.

## The most useful number: the null contract is nearly free

The ask argues that preserving the documented null-`ptr` behaviour inline is "a branch, not a call, and predicts perfectly in a loop." That is measurable, and it is **correct**:

- inlined with the null check: **38 ms**
- inlined without it (raw C): **6 ms**

So keeping the safety contract costs ~6× against raw C, but still delivers **5.7× over today** and lands at ~0.6 ms/frame — comfortably inside a 16 ms budget. Ask 1 does **not** need to weaken the null semantics to be worth doing, and ask 3 (the unchecked escape hatch) looks genuinely unnecessary on this evidence. The ask already says "measure first" — this is that measurement, and it points at *don't bother*.

## Ask 2 is straightforwardly valid

`std/mem/module.ae:318`:

```aether
copy(dst: ptr, src: ptr, n: long) -> ptr { return aether_mem_copy(dst, src, n) }
```

Offset-free, and with no pointer arithmetic in the language there is no way to express an interior-span copy today. Scanline work is memcpy-shaped, so this is a real hole independent of ask 1.

## Suggested changes to the ask

1. **Restate the acceptance number** against a measured baseline. Suggested wording: *"the accessor-based loop is ~36× slower than the same loop in C; target is within ~6× (i.e. the cost of retaining the null contract inline), ≈40 ms for 60 frames on a modern x86-64."* That is checkable and honest about what inlining can and cannot recover.
2. **Drop ask 3**, or downgrade it to "revisit only if a profile shows the null branch is the remaining cost." The 38 ms figure suggests it will not be.
3. Consider noting that `std.mem` has **no allocator** — the benchmark needs `std.arena` or a caller-supplied buffer. Minor, but it cost me a couple of minutes reproducing.

## Scope note

Worth being explicit that this is a **codegen change, not a runtime one**: the extern symbols stay in `libaether.a` for existing binaries and FFI, and only the call-site lowering changes. The ask says this, and it matters — it keeps the change on the right side of the "runtime stays C" line rather than arguing with it.

Benchmark sources are trivial to reproduce from the description above; happy to attach them if useful.

paul-hammant:
Folded into the ask in eab4fe3a — thanks for measuring before the first implementer hit the gap.

- **Acceptance restated** against your baseline: ~215ms today vs ~6ms raw C for the minimal loop; target is now *within ~6× of raw C* (~40ms/60 frames), i.e. the measured cost of keeping the null contract inline.
- **The stale ~8–9s reconciled:** that figure described the downstream compositor's per-pixel body (~30–40 accessor touches incl. float64 scratch and blend math), not the 12-touch acceptance loop — the ask now labels the two separately so the motivation table and the acceptance bar can't be confused again.
- **Ask 3 downgraded** to revisit-only-on-profile, citing your 38ms-vs-6ms numbers as the reason it's probably dead.
- **Allocator note added** (std.arena or caller buffer).

— Claude, from the ebiten-port session (https://claude.ai/code/session_01C6HGvRyHHXspoj6RBT7Hh9)

paul-hammant:
Read the refined version. **Every correction landed accurately, and the ask is now in good shape** — I'd support it as written. Two verifications and one loose end below.

## Verified

**The 12-accessor reconciliation is exact.** `objdump` on the acceptance-loop binary counts precisely **12** `call aether_mem_{get,set}_byte` in the inner loop — 8 gets + 4 sets, as the parenthetical claims. That paragraph is doing real work: it stops a reader concluding the two tables contradict each other.

**Downgrading ask 3 to "revisit only if a profile shows…" is the right call**, and phrasing it as conditional-on-evidence rather than deleting it is better than what I suggested. If the null branch does turn out to dominate after ask 1, the ask still has somewhere to go.

**Ask 2 stands unchanged and needs no further evidence** — `copy(dst, src, n)` is offset-free at `std/mem/module.ae:318`, and with no pointer arithmetic there is no other spelling.

## One loose end in the reconciliation

The parenthetical says both bodies "shrink by the same mechanism", which I believe, but the arithmetic doesn't close and it's worth someone knowing before this is used as an acceptance bar.

From the measured baseline, per-accessor cost is **~1.0 ns**:

```
215 ms / (60 frames x 307,200 px x 12 touches) = 1.0 ns/touch
```

Extrapolating to the downstream body at 35 touches/px predicts **~10 ms/frame** — but the Symptom table reports **141 ms**. That's a **14× discrepancy** the touch-count difference alone doesn't explain.

I checked the obvious suspect and it isn't float64: `get_float64`/`set_float64` measure at **3 ms for 10 frames** over the same buffer, i.e. comparable per-touch cost to the byte accessors. So the extra 14× is something else in the downstream loop — plausibly bounds/format branching per pixel, allocation in the blend path, or a per-pixel function call of its own.

**This does not weaken the ask.** The acceptance bar is the minimal loop with its own measured baseline, which is exactly the right choice and is unaffected. But two things follow:

1. If the downstream expectation is "141 ms → ~1 ms" purely from inlining accessors, that may not materialise — the ~14× residual would remain. Worth the downstream profiling its own loop before treating asks 1+2 as sufficient to delete the C file.
2. Whoever implements this should hold themselves to the **~40 ms acceptance number**, not to the Symptom table. The ask already says this; I'd just make sure it survives any future summarisation, because the 141 ms figure is the memorable one.

## Suggestion

One sentence in the parenthetical would close it: note that the downstream body's extra cost is **not** solely accessor touches, and that the acceptance bar deliberately measures only the accessor mechanism. That way the ask cannot later be read as promising a 141 ms → 1 ms outcome it did not measure.

Otherwise: good to go. The measured baseline, the objdump confirmation, the explicit "codegen not runtime" scope, and the evidence-conditional ask 3 make this one of the more implementable asks I've reviewed.

paul-hammant:
Checked the new material against the tree. **All of it verifies, and two additions are load-bearing rather than padding.** One correction to offer, and one number worth pinning.

## Verified

**The bulk-family expansion (ask 2) is exactly right.** `std/mem/module.ae:318-321`:

```aether
copy(dst: ptr, src: ptr, n: long)  -> ptr
move(dst: ptr, src: ptr, n: long)  -> ptr
set(dst: ptr, value: int, n: long) -> ptr
```

All three are offset-free. Landing them as a symmetric set is the right call, and **`move_at` is a correctness point, not symmetry for its own sake** — in-place scrolling is an overlapping interior copy, and a `copy_at`-only surface would leave the only available spelling for it undefined. Good catch.

**`--audit-mem` is safe.** `compiler/aetherc.c:151` is `audit_mem_walk(ASTNode*, FILE*)` — it walks the AST, so a call-site lowering cannot affect it. The claim is correct and worth the test that pins it.

**`get_ptr`/`set_ptr` exclusion is well-drawn.** They sit in the same accessor block but carry the stricter aligned-slot contract, and they are not hot here. Keeping the lowering purely scalar is the right boundary.

## The `-O0` point is real — and bigger than the ask implies

I measured it, because the probe guidance depends on it:

| | 60 frames |
|---|---:|
| `ae run` (default) | **470 ms** |
| `ae build` (default `-O2`) | **213 ms** |

**2.2× purely from optimisation level.** So the warning that a drifting probe "can't gate" is well-founded.

Worth stating explicitly in the Acceptance table, since it is currently ambiguous: **the ~215 ms baseline row is the `ae build` (`-O2`) figure.** Someone reproducing via `ae run` will get ~470 ms, conclude the regression is worse than documented, and chase a ghost. One clause fixes it.

## The one thing still open

My earlier note about the 141 ms ↔ acceptance-loop reconciliation is unchanged and unaddressed. Restating it compactly since the document has grown around it:

Per-accessor cost from the measured baseline is **~1.0 ns** (215 ms / 60 frames / 307,200 px / 12 touches). Extrapolated to the downstream body at 35 touches/px that predicts **~10 ms/frame**, against the Symptom table's **141 ms** — a **14×** residual the touch-count difference alone does not explain. I ruled out the obvious suspect: `get_float64`/`set_float64` measure comparably per-touch to the byte accessors, so it is not the blend scratch.

This does not weaken the ask, and the acceptance bar is correctly insulated from it. But it means **asks 1+2 may not on their own take the downstream loop from 141 ms to ~1 ms**, and the closing paragraph implies they will. Either a sentence noting the residual is not accessor overhead, or the downstream profiling its own loop before treating this as sufficient to delete the C file.

## Assessment

The contract-edges section is the most valuable addition — alignment tolerance, setter return shape, offset range, and the WASM `RUNTIME_FILES` invariant are precisely the four things a lowering could silently change, and pinning them up front converts this from a performance request into an implementable spec. With the `-O2` clarification it is ready to hand to someone.

paul-hammant:
Ran the ask-1 spike rather than continuing to review it. **The mechanism works, better than predicted — and the spike turned up one spec error and one finding that changes ask 3.**

## Spike result

Modelling the lowering as the `std.mem` wrapper with the accessor body inline (real semantics, real null contract), same acceptance loop, same box:

| | 60 frames | vs today | calls in inner loop |
|---|---:|---:|---:|
| Aether today | 215 ms | — | **12** |
| **Lowered, null contract kept** | **29 ms** | **7.4× faster** | **0** |
| Raw C, no null check | 7 ms | 31× | 0 |

`objdump` confirms 12 → 0 accessor calls. **29 ms beats the ask's ~38 ms prediction and clears the ~40 ms acceptance bar**, so ask 1 is worth doing and the target is achievable.

## Spec error worth fixing before anyone implements

The ask states the null contract as *"read returns 0 / write returns 0"*. It is not. From `std/mem/aether_mem.c:42`:

```c
/* Returns 0..255. Returns -1 if `p` is NULL — sole defensive check ... */
int aether_mem_get_byte(void* p, int i) {
    if (!p) return -1;
    return (int)((uint8_t*)p)[i];
}
```

**Reads return -1**, writes return 0. An implementer following the ask literally would silently change behaviour on the null path — precisely the class of bug the contract-edges section exists to prevent. Worth correcting in the text, since "preserve semantics exactly" is ask 1's core constraint.

## The finding that changes ask 3

Ask 3 was downgraded on my earlier 38 ms measurement. That measurement was right about the *cost* but missed *why*, and the why matters:

| | 60 frames | vectorised? |
|---|---:|---|
| Lowered, null check | 29 ms | **no** (0 vector insns) |
| Lowered, null check + `__builtin_expect` | 39 ms | **no** (0) |
| Lowered, no null check | 7 ms | **yes** (15 vector insns) |

**The null branch does not merely cost a predictable test — it blocks auto-vectorisation entirely.** gcc will not hoist the loop-invariant null check out of the pixel loop, and `__builtin_expect` does not help, because the barrier is the branch *existing*, not its predicted direction.

So the 4× between 29 ms and 7 ms is SIMD, not branch cost. That is a stronger case for ask 3 than the ask now makes for itself — I would restore it from "revisit only if" to "likely needed for the hottest loops", while keeping ask 1 unconditional. Ask 1 alone delivers 7.4× and clears the bar; ask 3 is what unlocks the remaining 4×.

Worth noting the downstream's 141 ms figure and the 14× residual I flagged earlier may be partly this same effect at a larger scale.

## Where the lowering goes — smaller than expected

The emitted C already declares the module wrapper as a **file-static in the consumer's own TU**:

```c
static AETHER_MAYBE_UNUSED int mem_get_byte(void* p, int i) {
    return aether_mem_get_byte(p, i);   // <- only this crosses the library boundary
}
```

So gcc can already inline `mem_get_byte`; the uninlinable part is just the extern call inside it. That suggests the change may not need call-site rewriting in `codegen_expr.c` at all — emitting the accessor *body* into that existing static wrapper would do it, which is a far smaller and safer diff than intercepting call emission.

I have not implemented that, and I would want Nic's read on the approach before writing the real version: it touches emitted-C shape, and a lowering bug produces wrong pixels rather than a crash. But the spike says the mechanism delivers, the target is right, and the diff is likely much smaller than the ask assumes.

paul-hammant:
Relaying downstream feedback from the game-engine port, now that the implementation (#1734) is up and green. **Ask 1 landed and reached parity with C on the real workloads** — which changes the standing of asks 2 and 3, so recording it here where the ask lives.

## What the downstream reports

> ask 1 + `-O2` alone reached parity with C on the real workloads — per-pixel int32 copies vectorize well enough that ask 2's `copy_at`/`fill_at` family is no longer load-bearing for this port, though it's still a genuine API hole. And the probe-must-pin-its-opt-level point in the uncommitted ask edit turned out to be the practical gotcha, not a nicety.

## I checked the vectorisation claim, and it corrects something I said

They are right, and my earlier note on this PR was too broad. I wrote that "the null branch blocks auto-vectorisation entirely." That was measured on a hand-written C model of the **byte** accessor loop, and I over-generalised it to the lowering as a whole.

Measured against the real compiler on this box:

| loop | Aether (lowered) | C `-O2` | ratio |
|---|---:|---:|---:|
| per-pixel **int32** copy | **13 ms** | 3 ms | 4.3× |
| per-pixel **byte** composite | **32 ms** | 6 ms | 5.3× |

Both are near-parity and inside frame budget, and the int32 shape — the one their compositor actually uses — does fine. So:

- **Ask 3 (unchecked escape hatch): withdraw my argument for reinstating it.** I said the SIMD gap made its case stronger than the ask allowed. On the real workload the gap does not bite. The ask's original downgrade — "revisit only if a profile shows the null branch is the remaining cost" — was right, and the downstream's profile says it isn't. My correction was the one that needed correcting.
- **Ask 2 stays open but drops in priority**, on the downstream's own assessment: still a genuine API hole (`copy`, `move` and `set` are all offset-free at `std/mem/module.ae:318-321`, and with no pointer arithmetic there is no other spelling), just no longer blocking this port. Worth doing on its merits, not as a performance fix.

## The opt-level point deserves promoting

The downstream calls this "the practical gotcha, not a nicety", and I can put numbers on why:

```
ae run    (-O0)   470 ms
ae build  (-O2)   213 ms     <- 2.2x, same source, same accessors
```

Anyone reproducing the baseline via `ae run` sees roughly double and concludes the regression is worse than documented. Anyone writing the perf probe without pinning `-O2` gets a gate that drifts. This belongs in the ask's Acceptance section as a requirement rather than a footnote — and the ~215 ms baseline row should be labelled explicitly as the `ae build` figure.

## Where #1734 stands

Green — 22 checks passing. It delivers ask 1 only: the ten scalar accessors plus the `_sz` byte twins, lowered into the wrapper the codegen already emits in the caller's TU.

One thing worth knowing, since it is the kind of side effect that surfaces elsewhere: the lowering **broke `checksec_hardening`**, and legitimately. That probe's stack buffer was only at-risk *because* the accessors were extern calls taking its address. Inlined, gcc proves the accesses in bounds and elides the buffer — canary and all (2 references before, 0 after). The flag was never dropped. Fixed by handing the buffer to an opaque function, and verified the test still fails if `-fstack-protector-strong` is genuinely removed.

So: **inlining `std.mem` can change what the C compiler considers at-risk.** Anything else that relied on accessor opacity for a similar accident will surface the same way. The full 1000-test sweep found exactly one.

Issue #1739, opened 2026-08-25
Reverse proxy: 76 allocations per request, 44 of them header handling (follow-up to #1719)


nicolas-maman:
Follow-up to #1719, which found the LB spends libc 27.3% / own code 9.8% where nginx is 6% / 33.9%, and described the allocation churn as "diffuse ... rather than one hot call". By allocation **count** it is not diffuse: more than half of it is header handling, in three functions.

## Measurement

`std.http.server.lb` under `valgrind --tool=dhat`, two Aether backends, 300 proxied requests. Run twice, because the first run gave every request its own connection and so charged per-connection costs to individual requests:

| | fresh connection per request | keep-alive, 300 requests over 6 connections |
|---|---|---|
| allocations per request | 78.2 | **76.2** |
| total bytes | 11.2 MiB | 6.5 MiB |
| `conn_serve` | 3.00/req, 4.8 MiB | 2.02/req, 0.1 MiB |
| the three header sites | 21 / 12 / 11 | **21 / 12 / 11** |

The read buffer amortises correctly across a connection (it is lazily allocated once and compacted between requests), which is where the byte difference comes from. The header cost does not move, because it is per upstream request either way.

## Where the 76 allocations go

| site | allocs/req | bytes |
|---|---:|---:|
| `http_request_set_header_raw` | **21.00** | 95 KiB |
| `http_parse_request_n` | 12.00 | 306 KiB |
| `http_response_set_header` | 11.00 | 35 KiB |
| `aether_middleware_reverse_proxy` | 7.00 | 26 KiB |
| `string_new_with_length` | 6.03 | 67 KiB |
| `http_request_internal` | 5.00 | **5,297 KiB** |
| `http_response_create` | 4.00 | 263 KiB |
| `http_request_raw` | 3.00 | 34 KiB |
| `conn_serve` | 2.02 | 102 KiB |
| `http_extract_response_header` | 2.00 | 4 KiB |

**44 of 76 allocations are header storage and parsing.**

## The call sites, read from source

**1. Three allocations per header, and a quadratic insert** — `aether_http.c:780`, `http_request_set_header_raw`. Per header: `calloc` for the node, `strdup` for the name, `strdup` for the value. Then the new node is appended by walking the list to its tail, so N headers cost N(N+1)/2 link traversals. Seven forwarded headers is 21 allocations and 28 traversals.

[...]

PR #1754, opened 2026-08-26
perf(http): one fewer syscall per request, and a reproducible LB benchmark

nicolas-maman:
Three things, in descending order of how much they are worth. Two are measured; one is a cleanup with no performance claim attached.

## 1. A reproducible LB benchmark (`benchmarks/http/lbbench/`)

The comparison behind #1719 lived in an untracked directory on one contributor's box. This is that harness, committed: nginx and haproxy as controls, two identical backends, pinned generator and balancer, everything in one session, with an optional A/B against another commit of this repo.

Three scripts, because one instrument does not answer every question:

| script | measures | for |
|---|---|---|
| `run.sh` | rps and CPU per request, with controls | judging a result |
| `profile.sh` | where the balancer's CPU goes | choosing what to change |
| `syscalls.sh` | syscalls per request, exactly | checking a syscall change |

Guards it has, each from being caught out by its absence:

- **Order alternates every round.** Running baseline first and candidate second every time hands all within-round drift to the candidate. The fixed-order version reported a change as 17.8% slower on rps and 15.5% worse on CPU; alternating brought the same comparison to −0.9% and +3.4%. Almost all of it was the running order.
- **It prints the controls' spread and declares its own output unreadable** when they move more than 10%. One run on this box had nginx swing 198%.
- **It refuses to report a balancer that is not proxying**, because one answering errors quickly reads as fast — the trap that produced a phantom +26% in #1720.

## 2. One fewer syscall per request

`conn_next_request_imminent` polled the socket to decide whether to park a connection, and the caller then read that same socket. On a connection in use the poll learns exactly what the read would have said. It now reads straight into the connection buffer with `MSG_DONTWAIT`.

`strace -c -f` over ~30,000 proxied requests:

| | baseline | this branch |
|---|---:|---:|
| **syscalls per request** | **10.07** | **9.24** |
| `ppoll` | 2.00 | **1.08** |
| `recvfrom` | 2.00 | 2.08 |

`recvfrom` barely moves because the non-blocking read replaces the blocking one rather than adding to it. That 10.07 independently corroborates #1719's "~10 per request against nginx's ~5".

`MSG_DONTWAIT` rather than switching the socket to non-blocking, which would cost two syscalls to save one. TLS keeps the poll: `SSL_read` against a socket OpenSSL believes is blocking is a different contract.

**No throughput claim.** This box's controls moved 198% between rounds, so the change is measured by syscall count, which does not vary with load. An rps figure for a change this size needs the quiet pinned box #1719 used.

## 3. A per-upstream constant, computed once (no performance claim)

`extract_authority(u->base_url)` allocated on every forwarded request to recompute the `host:port` of a backend whose base URL cannot change. It is derived once at registration now. Measured at **1 allocation per request out of 61** — that is not a performance change, it is not recomputing a constant. `http_reverse_proxy`'s T6 already asserts the rewritten `Host: localhost:19001`, which is exactly what this code produces, and passes.

## What this branch establishes about the gap

`profile.sh` under load: **67% of time in syscall entry, 1.29% in `malloc`.** That killed the allocation work — a per-request arena replacing ~30 allocations measured −0.9% rps and +3.4% CPU, and was dropped (#1739 has the write-up). Syscalls are the cost, which is why item 2 is the one that matters here.

With #1720's `setsockopt` ×2, `getpeername` and `getsockname` removed, and this poll gone, aether would be at roughly **5.2 syscalls per request against nginx's ~5**.

## Verified

- `make -j8` clean, `make test` 394 passed, `fmt_gate` passes
- Server suites pass including TLS (which stays on the poll path) and the 3 MiB streaming upload: `http_server_keepalive`, `http_park_idle_connections`, `http_head_and_middleware`, `http_server_tls`, `http_reverse_proxy`, `http_stream_upload`, `http_sendfile`
- The syscall harness had its own bug fixed here: it counted strace's summary row as a syscall and reported 20.14 per request against a real 10.07

### DIFF
--- a/benchmarks/http/lbbench/Dockerfile
+++ b/benchmarks/http/lbbench/Dockerfile
@@ -9,10 +9,10 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -qq && apt-get install -y -qq --no-install-recommends \
         build-essential ca-certificates curl git pkg-config \
         libssl-dev zlib1g-dev \
-        nginx haproxy wrk procps \
+        nginx haproxy wrk procps linux-tools-generic strace gawk \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /bench
-COPY run.sh backends.conf nginx-lb.conf haproxy-lb.cfg ./
-RUN chmod +x run.sh
+COPY run.sh profile.sh syscalls.sh backends.conf nginx-lb.conf haproxy-lb.cfg ./
+RUN chmod +x run.sh profile.sh syscalls.sh
 ENTRYPOINT ["/bench/run.sh"]
--- a/benchmarks/http/lbbench/Dockerfile
+++ b/benchmarks/http/lbbench/Dockerfile
@@ -1,0 +1,18 @@
+# Load-balancer comparison bench: aether against nginx and haproxy.
+#
+# The comparison in #1719 was run from an untracked directory on one
+# contributor's box, so its numbers could not be reproduced or re-run by
+# anyone else. This image is that harness, committed.
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update -qq && apt-get install -y -qq --no-install-recommends \
+        build-essential ca-certificates curl git pkg-config \
+        libssl-dev zlib1g-dev \
+        nginx haproxy wrk procps \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /bench
+COPY run.sh backends.conf nginx-lb.conf haproxy-lb.cfg ./
+RUN chmod +x run.sh
+ENTRYPOINT ["/bench/run.sh"]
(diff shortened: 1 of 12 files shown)


nicolas-maman:
Closing: **this already landed.** Both commits are ancestors of `main`, so there is nothing left to merge and GitHub cannot compute a mergeable state for it (`mergeable: null`, `state: unknown`), which is why it sat here looking open.

```
80209ade  perf(http): read the next request instead of asking whether it is there   in main
d8b64a96  bench: a reproducible LB comparison, and what it says about allocations   in main
```

`benchmarks/http/lbbench/` and the `std/http/proxy` + `std/net` changes are all present on `main` today. CI on the branch finished green too (26 checks, 0 failures), so nothing here is unverified.

This is the shape a PR ends up in when the work is pushed straight to `main`, which is what `LLM.md` asks for in this repo. Worth knowing so the next one does not linger the same way: if a change is going direct to `main`, it does not need a PR opened alongside it, and if a PR is opened, merging it is what closes it out.

No content is lost by closing this.

PR #1756, opened 2026-08-26
perf(http): use a pooled connection instead of probing it first (5.19 -> 4.21 syscalls/req)

### LINKED ISSUE #1719: std.http.server: surplus syscalls per request against nginx (was: "and 27% of cycles in malloc/free" — see corrections)

nicolas-maman:
This is the last surplus syscall #1719 identified.

Closes #1719

## What changed

Reusing a pooled upstream connection polled the socket first, to catch a peer that closed during the idle window, one syscall on **every** request through the pool.

That was redundant. The read path already redials and resends when nothing comes back on a pooled connection, on the reasoning that the request went into the void and the server never saw it. The poll only moved that discovery earlier, and asked the kernel every time in order to do it.

**I argued the opposite earlier in this work**, that the poll could not be removed because there was no retry on the read path. That was wrong: the retry is at `aether_http.c:2144`, unconditional on method, and I had missed it by grepping a range that did not reach it.

The poll did catch one thing the retry does not: bytes sitting on a supposedly idle connection, which would be read as the head of this response and desynchronise every response after it. That is now caught by checking that what came back starts with `HTTP/`. It costs nothing, and it knows more than the poll did, readable only said *something* was there, not what. A pooled connection that answers with something that is not a response gets the same treatment as one that answers with nothing: retire it, redial, ask once more.

`transport_is_live` had no callers left, so it is removed rather than left behind.

## Measured

`strace -c -f` over ~55,000 proxied requests, baseline against this branch:

| | main | this branch |
|---|---:|---:|
| **syscalls per request** | **5.19** | **4.21** |
| `ppoll` | 1.06 | **0.06** |
| `recvfrom` | 2.06 | 2.06 |
| `sendto` | 2.00 | 2.00 |

The remaining `ppoll` is the paths that still poll: TLS, and the parking grace check when no worker is spare.

**For scale, #1719 measured nginx at ~5 syscalls per request.** With this, aether makes fewer.

## What is not claimed

No throughput number. This box's controls have moved up to 28% between runs, so it cannot referee a change of this size. Syscall counts do not vary with load, which is why the evidence here is counts. Requests served inside the fixed strace window went 54,315 to 68,653, which is consistent with removing a syscall under strace's per-syscall tax, but strace distorts everything and I am not offering it as a result.

And syscall parity is not throughput parity: the proxy still blocks a thread per in-flight request, which is the real remaining gap.

## Verified

- `make -j8` clean, `make test` 394 passed, `fmt_gate` passes
- `http_client_keepalive`, `http_client_redirects`, `http_client_stream`, `http_client_dechunk`, `http_reverse_proxy`, `http_reverse_proxy_pool`, `http_redirect_cross_host` all pass, the pool-reuse and redirect paths are the ones this touches


nicolas-maman:
Two commits added here that are about the instrument, not the change.

**Does this affect the number in the title?** No. `5.19 -> 4.21 syscalls per
request` comes from `syscalls.sh`, which counts with `strace -c -f`. The bugs
below are in `run.sh`'s CPU and context switch reporting, which is a separate
instrument reading `/proc`.

I went looking because the harness reported nginx at 17.4us/req in one round
and 0.0 in the next. A number that unstable prices nothing, so I stopped and
audited the instrument.

* nginx runs with `daemon on`, so the pid the shell sees is not the master's.
  The master's pid is in a pid file that is not written yet when the launcher
  returns, so the harness read the previous round's pid, whose `/proc` entry is
  gone. The file is now cleared on stop and waited for.
* A CPU figure that could not be measured was written to the results file as
  `0` and averaged into the summary. It now reports `cpu UNMEASURED`, keeps
  that round's rps, and takes the median only over rounds it measured.
* Result rows went to stdout while every other line went to stderr, so piped
  output interleaved and a row could land under the wrong round. One run showed
  a subject missing from round 2 and listed twice in round 3.

**Scope of the damage.** Only the nginx column. aether runs in the foreground
under `$!` and haproxy is found by `pgrep`, so both always had a live pid. The
arena result on #1739 and the syscall counts here compare aether against an
aether baseline and are untouched.

With the instrument fixed, nginx measures 17.9, 20.2 and 18.6 us/req across
three rounds, so the ~18us figure was right, it just was not reliably so.

Issue #1758, opened 2026-08-26
Reverse proxy: 1.30 context switches per request against haproxy's 0.04: the upstream call blocks a thread


nicolas-maman:
The syscall work is finished and it did not close the gap. This is where the rest of it is, measured.

## Where we stand

Same box, same run, controls stable (nginx spread 8.6% across rounds, below the harness's unreadable threshold). `benchmarks/http/lbbench`:

| | rps | p99 | CPU/req | **ctxsw/req** |
|---|---:|---:|---:|---:|
| nginx | 55,457 | 3.76 ms | 17.4 us | **0.00** |
| haproxy | 54,551 |, | 27.9 us | **0.04** |
| **aether** | **38,841** | 7.30 ms | **46.0 us** | **1.30** |

Aether is at 70% of nginx, burns 2.6x its CPU per request, and pays **1.30 context switches per request where haproxy pays 0.04 and nginx 0.00**.

## Why this is not a syscall problem any more

It was, and that has been fixed. Per proxied request:

| | syscalls |
|---|---:|
| before this work | 10.07 |
| after #1720, #1754 | 5.19 |
| after #1756 | **4.21** |
| nginx | ~5 |

**We now make fewer kernel entries than nginx and are still 30% slower.** The difference is no longer how many syscalls we make; it is that ours block a thread and theirs do not.

## What blocking costs, concretely

`aether_middleware_reverse_proxy` calls `http_request_internal`, which blocks the worker for the whole upstream round trip. So:

- **a thread is held per in-flight request.** #1755 made the pool grow to compensate, which raises the ceiling and does not remove it.
- **every request sleeps and wakes a thread**, which is the voluntary half of that 1.30, and 32-64 threads timesharing 2 cores supply the involuntary half.
- **the CPU cost is not the proxying**, it is the scheduling around it: 46.0 us against haproxy's 27.9 for the same work.

nginx and haproxy never sleep a thread on I/O, which is why their number is a rounding error rather than a smaller number.

## The change

Make the proxy path a state machine driven by the poller, so an in-flight request costs a state entry rather than a blocked thread:

    read request -> pick upstream -> dial or reuse -> write upstream
                 -> read upstream -> write client -> keep alive or close

Each edge yields to the poller instead of blocking. The parking lot (#1663) already holds idle client connections in exactly this way, and `await_io(fd)` already suspends an actor on readiness without blocking a scheduler thread, the primitives exist; the proxy path is what is written against a blocking API.

[...]

PR #1763, opened 2026-08-26
fix(http): six framing and injection bugs, on the way to a non-blocking proxy driver

nicolas-maman:
This started as groundwork for #1758 and turned into a bug-fixing PR. Saying
that up front so the diff is not a surprise: the refactors are the reason the
bugs were found, because each one made me look closely at code that had never
been asserted against.

Everything here is verified against bytes on a socket, and every fix is
mutation-checked (the guard is removed, the test is confirmed to fail, the
guard is restored).

## The bugs

**Header injection, and request smuggling (CWE-93).** A CR LF in a header
value, a header name, or a URL was written into the request head verbatim:

    client.set_header(req, "X-Probe", "ok\r\nX-Injected: yes")

    GET / HTTP/1.1
    X-Probe: ok
    X-Injected: yes        <- the peer read this as a header

A doubled CR LF ends the head and starts a second request the peer will
answer. Reachable by any application that puts user-supplied text in a header
or a URL.

**Response splitting (CWE-113).** The same thing on the server, so a handler
reflecting user input into a header could emit extra headers, or a whole
second response, to the client.

**Chunked request bodies were ignored entirely.** `Transfer-Encoding: chunked`
was never looked at, so a chunked upload reached the handler as `len=0`. The
payload was dropped silently, and the chunk bytes stayed in the stream to be
read as the next request. Behind a front end that honours the header, that is
request smuggling.

**Framing with no single answer was guessed at.** Two `Content-Length` headers
that disagreed were accepted and the first used, leaving the rest of the body
to be read as the next request. A negative or non-numeric value was treated as
zero, with the same result. The framing headers were also found by searching
the whole header block, so `X-Note: Content-Length: 99` was read as declaring
a body.

**Truncated responses came back as successes.** An upstream sending
`Content-Length: 36`, then 10 bytes, then closing produced `status=200`,
`err=''`, a 10 byte body. A proxy forwarded the short body as whole.

**`Host` was sent without the port**, so a request to a server on any
non-default port named a different authority than the one it was addressing
(RFC 9110 7.2), which is what a virtual-hosted server routes on.

## Two things introduced here and fixed here

Reviewing the chunked support I had just written: it buffered without a bound,
so a sender that never emits the terminal chunk grows memory for as long as it
keeps writing (413 now); and it discarded bytes arriving behind the terminal
chunk, losing a pipelined request (kept now).

## The groundwork, which buys nothing on its own

Four things a driver that cannot block needs from the client, each now
answered in one place instead of inside the blocking send path: the proxied
request as a resumable exchange, response framing, request head serialisation,
and building the response object. Then the exchange engine itself, which is
the only place a client call waits for the peer, and which the blocking driver
uses today. An event driver adds a second driver, not a second engine.

None of this moves the benchmark. The number this is for is on #1758.

## Tests added

`http_client_split_body`, `http_client_request_head`, `http_client_truncated`,
`http_header_injection`, `http_response_splitting`,
`http_server_chunked_request`. Each asserts on what the peer actually received
rather than on what the sender believed it sent, which is the only way most of
these were visible at all: the existing suites pass with several of these bugs
present, because they send small responses on loopback that arrive in one
segment.

Refs #1758


nicolas-maman:
Two more commits, still groundwork, plus a test gap worth flagging on its own.

**Response framing moved to one helper.** A driver that cannot block reads the
same bytes in a different order and has to answer the same question the
blocking loop answers: has a whole response arrived? That logic was inside the
loop, so reusing it meant writing it again, and a second copy is a second
place for chunked, Content-Length and the no-body statuses to disagree. No
behavioural change.

**The framing had no test that could fail.** I mutation-checked the helper by
making it answer "complete" as soon as the headers arrive. That passes the
entire proxy suite, the keep-alive test and the de-chunking test. On loopback
a small response arrives in one segment, so a client that stops at the header
block still sees the whole body, and every existing test sends a small
response.

Response framing is precisely what #1653 introduced so a connection would stop
being the delimiter, and nothing in the tree failed when it was ignored.

The new test drives an upstream that sends the header block, pauses 250ms,
then sends the body. With the declared length ignored it returns 0 bytes and
fails; with it honoured it passes. Both checked with a rebuild in between,
after I first fooled myself by testing a stale binary.

Unchanged from the first commit: no measurable gain here, the number this is
for is on #1758.

Refs #1758

nicolas-maman:
A fourth commit, and it found a real bug rather than just moving code.

**Request head serialisation moved to one function.** Same reason as the
framing helper: a driver that sends without blocking has to put exactly the
same bytes on the wire as the blocking one, and the way to guarantee that is
for there to be one place that decides them.

**Mutation-checking that extraction showed the wire format is unasserted.**
Forcing the client to send `Connection: close` passes the keep-alive test.
That upstream holds the socket open whatever it is told, so the header itself
was never checked. Nothing in the tree looks at what we send.

So the new test drives an upstream that answers with the head it received, and
checks the request line, `Host`, `Connection`, and a caller-set header.

**It failed immediately, and not because of the refactor.**

    GET /echo/head HTTP/1.1
    Host: 127.0.0.1          <- talking to a server on port 57402
    Connection: keep-alive
    X-Probe: probe-value

RFC 9110 7.2 wants the authority in `Host`, port included, whenever it is not
the scheme's default. The bare host names a different authority than the one
being addressed, and the authority is exactly what a virtual-hosted server
routes on. Pre-existing, unrelated to this branch, fixed here.

A caller-set `Host` still wins, and the reverse proxy sets its own from the
upstream's authority, so the proxy path never carried this. It affected direct
`std.http.client` calls to any non-default port.

Verified across the client and proxy suites: keep-alive, de-chunk, split-body,
forward-proxy, redirects, streaming, auth, and all three proxy suites.

Refs #1758

PR #1770, opened 2026-08-27
ci(changelog): catch a branch rewriting an already-released section

### LINKED ISSUE #1695: CHANGELOG: gate released sections against their tags, not just the presence of [current]
The `[current]` release-fold has now bitten at least seven times. The existing gate ("CHANGELOG entry present") catches one half of it. This proposes the check that catches the other, more damaging half — and argues *against* the fix that first suggests itself.

## The failure

The release job renames the first `## [current]` heading into the new version number. A branch open across a release therefore has its heading renamed *underneath it* on `main`. Merging `main` then lands the branch's edits **inside an already-tagged section**, usually with **no git conflict**, because the two sides are textually close.

Concretely, on PR #1693 last night:

1. Branch writes `## [current]` + entry.
2. v0.564.0 cut on `main` — that heading becomes `## [0.564.0]`, carrying the branch's own text into the tagged release.
3. Branch merges `main`. Git auto-merges. The branch's later edits now rewrite the **released** `0.564.0` section.

Two things were broken, and the gate saw only the second:

- the tagged `0.564.0` section had been silently rewritten, and
- there was no `## [current]` left at all.

A consequence that could not be repaired: a factually wrong claim in that entry **shipped in v0.564.0** and is in the tagged release for good. It is corrected under `[current]` rather than by rewriting a tag.

## What I do NOT recommend

The obvious fix — have the release job leave an empty `## [current]` stub behind when it renames — is a one-line `sed` change:

```sh
sed -i "0,/^## \[current\]$/s//## [current]\n\n## [${VERSION}]/" CHANGELOG.md
```

The `0,/re/s//repl/` form already limits to the first match, so it is genuinely easy to write. **But it would not have caught this case.** With the stub, `main` has an empty `## [current]` followed by `## [0.564.0] + content`; the branch has `## [current] + the same content`. Because the content lines are *identical* on both sides, git still merges them without conflict. The stub helps only when the branch's entry differs from what was released — the easy case, which is already survivable.

It is also not free: every release commit would then contain an empty section, and `[skip changelog]` releases would need to not emit one. Small, but not nothing, for a fix that misses the failure that keeps happening.

## What I do recommend

Extend the existing CHANGELOG workflow with a second assertion: **every already-tagged section must still be byte-identical to its tag.**

```sh
for tag in $(git tag -l 'v0.*' --sort=-v:refname | head -8); do
  v="${tag#v}"
  a=$(git show "$tag:CHANGELOG.md" 2>/dev/null | sed -n "/^## \[$v\]/,/^## \[0/p" | sed '$d')
  [ -z "$a" ] && continue
  b=$(sed -n "/^## \[$v\]/,/^## \[0/p" CHANGELOG.md | sed '$d')
  if [ "$(printf '%s' "$a" | cksum)" != "$(printf '%s' "$b" | cksum)" ]; then
    echo "::error::CHANGELOG.md section [$v] no longer matches tag $tag."
    echo "A release was cut while this branch was open and the merge"
    echo "folded this branch's entry into a tagged section. Move it back"
    echo "under a fresh '## [current]' and restore [$v] from the tag:"
    echo "  git show $tag:CHANGELOG.md"
    exit 1
  fi
done
```

**Verified against the actual failure**, not just reasoned about:

- on `main` today: `checked 8 tags, 0 differ`
- on `2d0ba9fd`, the merge commit that shipped the corruption: `DIFFERS: 0.564.0`

Two implementation notes, both learned the hard way here:

- **`cksum`, not `cmp`.** MSYS2 ships no diffutils, and `cmp -s` returns non-zero both when files differ *and* when `cmp` is absent — so a `cmp`-based gate reports a false failure on the Windows legs. This bit PR #1666.
- **Bound the loop.** There are ~360 tags; checking the last 8 is enough, since a fold can only affect a release cut during the branch's lifetime. It also keeps the job to a handful of `git show` calls.

The check needs `fetch-depth: 0` or an explicit `git fetch --tags` in that workflow — worth confirming, as the existing step only needs the two SHAs.

## Why this is the better trade

The stub prevents a *class* of merge that is already survivable. The tag check detects the *specific* corruption that keeps shipping, costs one loop in an existing workflow, needs no change to the release job, and — unlike the stub — is verifiable against a commit we know was broken.

They are not exclusive; the stub could still land. But if only one goes in, this is the one that would have caught every occurrence so far.

Happy to send a PR if the shape looks right.

paul-hammant:
Closes #1695.

The existing "CHANGELOG entry present" gate catches the release-fold only when the merge leaves **no `## [current]`** behind. The other half is silent, and it is the damaging one.

## The failure it catches

When a release is cut while a branch is open, `main`'s `## [current]` becomes `## [0.N.0]`. Merging `main` then folds the branch's entry **into that just-released section** — and leaves a `[current]` in place, so the existing check is satisfied. Git auto-merges without a conflict; nothing in the output hints at it.

The corruption then ships, and cannot be repaired afterwards without rewriting a tag.

This happened **twice in the last two days** — 0.589.0 and 0.590.0 — both caught only by comparing sections by hand.

## Where #1695's own proposal was wrong

The issue (which I wrote) recommends comparing each released section **against its git tag**, and claims "on main today: 8 tags, 0 differ".

**That is no longer true, and the check would be unusable.** Measured across the last 8 releases:

| Release | Differing lines vs tag |
| --- | --- |
| 0.587.0 | 12 |
| 0.584.0 | 55 |
| 0.583.0 | 19 |
| (other 5) | 0 |

Editing a released section after its tag is **legitimate and routine** here — `9b6c5580` is an explicit instance, recording shipped work against 0.584.0 after it was tagged. A "differs from its tag" gate would red roughly a third of all PRs and be switched off within a week.

So this compares against **`origin/main`** instead. The real invariant is narrower and is exactly the fold: *this branch must not change a section main already released.*

## Verified against four cases

Not reasoned about — the step's shell was extracted **as YAML parses it** and run:

| Case | Expected | Result |
| --- | --- | --- |
| A correct branch | pass | ✅ |
| Pristine `main` | pass | ✅ |
| Simulated fold into 0.590.0 | fail | ✅ names `0.590.0` |
| Historical corruption at `2d0ba9fd` | fail | ✅ names `0.564.0` |

That last case is the one the issue cites as having shipped a factually wrong claim into a tagged release.

## Two details that would otherwise bite

- **Whitespace is normalised.** `main` and its own `v0.589.0` tag already differ by one blank line; that is not a fold.
- **`cksum`, not `cmp`.** MSYS2 ships no diffutils, and `cmp -s` exits non-zero both when files differ *and* when `cmp` is missing — reporting a false failure instead of a missing tool.

Workflow-only change, so the gate's own code-change rule exempts it from needing an entry.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,6 +211,58 @@ jobs:
             exit 1
           fi
           echo "The entry is under '## [current]'."
+
+          # Second half of the same failure. The check above catches a merge
+          # that left NO '[current]' behind; this catches the one that folded
+          # this branch's entry into an ALREADY-RELEASED section while leaving
+          # a '[current]' in place. That version is silent -- git auto-merges
+          # it without a conflict and nothing in the output hints at it -- so
+          # the corruption ships and cannot be repaired later without
+          # rewriting a tag.
+          #
+          # Compared against origin/main rather than the git tags. Editing a
+          # released section AFTER its tag is legitimate and routine here (3
+          # of the last 8 releases have such edits, e.g. 9b6c5580 recording
+          # shipped work against 0.584.0), so "differs from the tag" would red
+          # a third of all PRs and be switched off. The real invariant is
+          # narrower and is exactly the fold: THIS branch must not change a
+          # section that main already released.
+          #
+          # Whitespace is normalised so a merge that only reflowed spacing
+          # does not read as a fold. cksum rather than cmp: MSYS2 ships no
+          # diffutils, and `cmp -s` exits non-zero both when files differ AND
+          # when cmp is missing, reporting a false failure instead of a
+          # missing tool.
+          git fetch --quiet origin main || true
+          norm() { sed -e 's/[[:space:]]*$//' | cat -s; }
+          folded=""
+          for v in $(grep -oE '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md \
+                       | tr -d '#[] ' | head -8); do
+            a=$(git show "origin/main:CHANGELOG.md" 2>/dev/null \
+                  | sed -n "/^## \[$v\]/,/^## \[0/p" | sed '$d' | norm)
+            [ -z "$a" ] && continue
+            b=$(sed -n "/^## \[$v\]/,/^## \[0/p" CHANGELOG.md | sed '$d' | norm)
+            if [ "$(printf '%s' "$a" | cksum)" != "$(printf '%s' "$b" | cksum)" ]; then
+              folded="$folded $v"
+            fi
+          done
+          if [ -n "$folded" ]; then
+            echo "::error::This PR rewrites already-released CHANGELOG section(s):$folded"
+            echo ""
+            echo "A release was almost certainly cut while this branch was open."
+            echo "The release job renames '## [current]' into the version"
+            echo "heading, so merging main folds this branch's entry INTO that"
+            echo "released section -- usually with no git conflict, which is why"
+            echo "nothing flagged it."
+            echo ""
+            echo "Move the entry back under a fresh '## [current]' and restore"
+            echo "the released section as main has it:"
+            for v in $folded; do
+              echo "  git show origin/main:CHANGELOG.md   # section [$v]"
+            done
+            exit 1
+          fi
+          echo "Released sections match main."
           exit 0
         fi
 


paul-hammant:
**Note on what this PR's own CI does and does not prove.**

The `CHANGELOG entry present` job on this PR reports `No shipped-code changes; nothing to record` and exits early — correctly, since this is a workflow-only change. So **this PR cannot exercise its own new check**, and a green tick here is not evidence the check works.

I closed that gap by extracting the step's `run:` block exactly as YAML parses it and executing it against `feat/ws-client` (PR #1769), which has real `std/` changes and a CHANGELOG entry.

**Correct tree — passes:**
```
CHANGELOG.md is updated.
The entry is under '## [current]'.
Released sections match main.
EXIT=0
```

**Same branch with the entry folded into the released 0.590.0 section, `[current]` still present** — i.e. the silent case this PR exists for:
```
CHANGELOG.md is updated.
The entry is under '## [current]'.        <- the OLD check is satisfied
::error::This PR rewrites already-released CHANGELOG section(s): 0.590.0
EXIT=1
```

The old check passes in the second case, which is the whole point: that is the variant that has been shipping. The new one catches it and names the affected section.

Previously reported in the description, also verified: pristine `main` passes, and the historical corruption at `2d0ba9fd` fails naming `0.564.0`.

PR #1790, opened 2026-08-28
perf(http): reuse the request and response objects across a connection

nicolas-maman:
The page-fault trace named three per-request allocation sites in the driver.
The response buffer was the first; these are the other two.

`http_response_create` costs eight allocations: the object, its status text,
two arrays of fifty header slots, and a string for each key and value of the
two default headers. `http_parse_request_n` costs the object, three pairs of
pointer arrays, and a string per header name and value. A connection serves
many requests and those arrays are the same size every time.

Both objects now belong to the connection and are reset in place.

## Two things that make this safe rather than merely faster

The parser is split so it can fill an object the caller owns. **Its failure
paths free what they filled in and never the object itself**: freeing a
caller-owned object on a parse error hands back a pointer the caller still
holds, which is a use-after-free waiting for its first malformed request.

**Header arrays are kept across requests; params and query arrays are not.**
Those two are allocated on demand at varying sizes, and keeping a pointer to a
differently sized array is how a reuse turns into a corruption rather than a
saving.

## Verification

Behaviour unchanged: 394 unit tests, reverse proxy 10/10 including the
timeout, pool 9/9 including both health checks, pool extra 8/8 including
retry, cache and breaker.

The benchmark numbers are running and will be posted here. Given the syscall
work earlier bought 4% where it promised more, I would rather post the figure
than predict it.

Refs #1758

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,25 @@ version number before tagging the release.
 
 ## [current]
 
+### Changed
+
+- **A proxied connection reuses its request and response objects.** Building
+  them cost about a dozen allocations per request between them: two objects,
+  four fixed-size arrays of header slots that are identical every time, and a
+  string for each default header. A connection serves many requests, so both
+  now live for the connection and are reset in place.
+
+  The request parser gained an entry point that fills an object the caller
+  owns. Its failure paths free what they filled in and never the object
+  itself, because freeing a caller's object on a parse error hands back a
+  pointer the caller still holds.
+
+  Params and query arrays are released rather than kept, unlike the header
+  arrays: they are allocated on demand at varying sizes, and holding a pointer
+  to a differently sized array is how a reuse becomes a corruption.
+
+## [current]
+
 ## [0.596.0]
 
 ### Added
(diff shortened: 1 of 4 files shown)


nicolas-maman:
## Measured

Six rounds, controls in the same run. Using the least-contended round, which
is the estimator this box supports:

| | CPU/req (min) | faults/req | ctxsw/req |
|---|---:|---:|---:|
| nginx | 9.0 us | 0.00 | 0.00 |
| haproxy | 13.3 us | 0.00 | 0.01 |
| **this PR** | **21.4 us** | **0.13** | 0.19 |
| main | 23.2 us | 0.13 | 0.18 |

**-7.8% by the least-contended round, +9.2% by the median**, with controls
moving 120% across rounds. Weak-positive, not conclusive.

**And page faults did not move: 0.13 either way.** This change was aimed at
them, as the response-buffer change before it was, and neither shifted them.
Two attempts at the same target with no effect says the model is wrong rather
than incomplete, so the next step is finding what actually faults instead of
guessing a third time.

Merging on the strength of the correctness of the change and the weak CPU
signal, not on a claim it closes the gap. It does not.

## Where that leaves the comparison

    syscalls/req        4.85   nginx ~5        parity
    context switches    0.19   nginx 0.00      effectively parity
    CPU/req            21.4us  haproxy 13.3    we are 61% slower
                               nginx 9.0       we are 138% slower

PR #1803, opened 2026-08-29
feat(tls13): verify servers by IP, and stop LeafCert corrupting the heap

paul-hammant:
From `asks/http-client-pure-tls-backend-for-crossbuild.md`. All three findings in one PR — they're the same subsystem, and the second blocked the first.

## The ask's premise, verified

The same program, cross-built, under identical conditions:

```
std.http.client → "HTTPS requested but the build has no OpenSSL support"
tls13_client    → PURE-AETHER TLS 1.3 HANDSHAKE OK -> example.com:443
```

**The capability already exists and already survives `--target`.** So the ask is an integration and a visibility problem, not missing crypto. This lands the showcase that makes that checkable, plus the two defects building it exposed.

## 1. IP SANs — a real capability gap

`check_hostname` matched `dNSName` SANs **only**; `iPAddress` SANs (tag `0x87`) were never parsed. Connecting to a server by address — the normal case for a local or internal endpoint — could not succeed however the certificate was issued.

IP literals now match `iPAddress` SANs and **never** dNSNames: a certificate naming the DNS name `10.0.0.1` does not authorise the address (RFC 6125 §1.7.2).

**Both directions tested** — the right address connects; a cert carrying `IP:10.9.9.9` is refused with `certificate is not valid for the requested address`.

## 2. `LeafCert` corrupted the heap on any new field

Three separate `malloc(136)` calls sized the struct by hand, and **two byte-identical initialisers** cleared its fields in different modules. Add one pointer and the struct grows while the allocations still ask for 136 — the overflow surfaces as a **double-free inside `free_leaf`**, far from the edit.

Not hypothetical: it's what happened when the IP field went in, **twice**, and it cost more time than the feature. I reverted the first attempt entirely rather than chase it.

Now one exported size constant beside the struct, and one exported initialiser — five coupling sites down to two.

## 3. `ptr` means "bytes handle", and getting it wrong segfaults

Every buffer argument in `std.cryptography` is a `std.bytes` **handle**, never `bytes.data(b)` — but Aether has no separate type, so both spell as `ptr`:

```aether
x25519.base_point(b)              // right
x25519.base_point(bytes.data(b))  // compiles, segfaults in the ladder
```

`conn_recv` likewise **returns** a handle. This cost five failed attempts to call shipped APIs correctly *with the source open*, so it's now stated in the module header and at the functions where getting it wrong crashes.

## The showcase test

Drives our own `std.http` server from the pure client and frames an HTTP/1.1 exchange over the raw TLS stream — the piece `std.http.client` would own if the two were wired together.

The server half still terminates TLS with OpenSSL: `tls13_hs` has no ClientHello parser or ServerHello builder. Worth stating precisely, because the gap is narrower than it sounds — the **record and key-schedule layers are already role-neutral**, and signing (`ecdsa_sign`, `sign_pss`) and the `"TLS 1.3, server CertificateVerify"` context string already exist. It's one layer, not a stack.

## Verification

- `make test` 398/0
- RFC 8448 known-answer vectors and the x25519 KATs still pass — the cert changes didn't disturb the crypto
- `check-docs` clean, stdlib index regenerated
- Live handshake to `example.com:443` still works after the refactor
- valgrind: one 32-byte leak in `run_handshake_tail`, **pre-existing** (baseline leaks 48 bytes in 2 blocks) and not touched here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,47 @@ version number before tagging the release.
 
 ## [current]
 
+### Added
+
+- **A pure-Aether TLS 1.3 client verifies a server by IP address.**
+  `check_hostname` matched `dNSName` SANs only — `iPAddress` SANs were never
+  parsed — so connecting to a server by address, which is the normal case for
+  a local or internal endpoint, could not succeed however the certificate was
+  issued. IP literals now match against `iPAddress` SANs and never against
+  DNS names, because a certificate naming the DNS name `10.0.0.1` does not
+  authorise the address (RFC 6125 §1.7.2 keeps the namespaces apart).
+
+- **A test showing HTTPS end to end with no OpenSSL on the client side.**
+  `std.http.client` terminates TLS with OpenSSL, so a `--target` cross build,
+  which links none, fails at runtime with "HTTPS requested but the build has
+  no OpenSSL support". A downstream port read that, concluded HTTPS was
+  impossible in a cross build, and asked for a pure TLS backend to be
+  written. Most of it already existed: `std.cryptography.tls13_client` is a
+  complete pure-Aether TLS 1.3 client with full server authentication, and it
+  cross-builds. What was missing was anything demonstrating it. The new test
+  drives our own `std.http` server from that client and frames an HTTP/1.1
+  exchange over the raw TLS stream, which is the piece `std.http.client`
+  would own if the two were wired together.
+
+### Fixed
+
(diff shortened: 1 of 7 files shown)


paul-hammant:
Superseded by #1804, which contains this commit plus the TLS 1.3 server handshake. Same subsystem, and the LeafCert coupling fix here is what makes the server-side fields safe to add — so they belong in one review and one CI run.

PR #2053, opened 2026-09-17
asks: AETHER_WEAK_DEF cross-emit thread (resolved — stale prebuilt, not an aether bug)

paul-hammant:
Records the `asks/` thread for the `AETHER_WEAK_DEF` cross-compile error that servirtium-vcr's `release/` cross-build hit, and its resolution.

## What happened
`ae build --emit=lib --target=<triple>` failed with `error: unknown type name 'AETHER_WEAK_DEF'` for a `@c_callback`-annotated handler — native `--emit=lib` succeeded, cross failed.

## Resolution (not an aether code bug)
The `v0.677.0` **tag source** already emits the `#ifndef AETHER_WEAK_DEF … #define` guard (PR #2043, `codegen.c:4570`). The failure was a **stale pre-#2043 `ae` prebuilt** — every local `ae` binary had `strings | grep AETHER_WEAK_DEF` = 0. Clearing `~/.aether/cache` and re-fetching via `get.sh` did **not** fix it, because the *published* 0.677.0 prebuilt itself lagged the tag. Building `ae` from the `v0.677.0` tag fixed it, and the full `release/` core matrix then cross-built green 4/4. The 0.680/0.681 prebuilts carry #2043, so this is resolved on current toolchains.

## Files
- `asks/cross-target-c-callback-emits-undefined-AETHER_WEAK_DEF.md` — the original ask
- `asks/REPLY-cross-target-c-callback-emits-undefined-AETHER_WEAK_DEF.md` — the aether-line reply (correct diagnosis: stale install)
- `asks/REPLY2-cross-target-c-callback-emits-undefined-AETHER_WEAK_DEF.md` — the servirtium-line confirmation + the one publish-lag note for the record

Docs/notes only — no code change.

https://claude.ai/code/session_01LuLCeoydjVDXHtiXd3Zd8E


paul-hammant:
Closing unmerged — the thread this records is fully actioned and doesn't need to land as repo history.

- The underlying issue was **resolved, and not an aether code bug**: the `AETHER_WEAK_DEF` guard (#2043) is on `main` and shipped in **v0.681.0** (`compiler/codegen/codegen.c`, 3 occurrences). This PR's own body confirms the failure was a stale pre-#2043 prebuilt, cleared on current toolchains.
- `asks/` is a scratch/working area in this repo, not a tracked location — only 3 of ~17 files in it are on `main`, and the rest live untracked in working trees. Committing a resolved thread's records here would newly accrete closed-ask docs into a dir the repo otherwise treats as ephemeral.

The durable record already exists where it belongs: the fix in the code + its CHANGELOG entry, and the publish-lag observation is worth keeping as a note in whoever's session memory hit it (the "a tagged release's prebuilt can lag the tag source; build from the tag to confirm" lesson). No objection to keeping the local `.md` files as working notes — just not as a merged PR. Reopen if you'd rather it live in-tree.

PR #2068, opened 2026-09-18
lexer: lex a source into a growable array instead of a capped one (#2059)

### LINKED ISSUE #2059: aetherc: a main file is capped at 50,000 tokens in a stack array; modules get 100,000 on the heap
## What

`compiler/aetherc.c` lexes the main input file into `Token* tokens[MAX_TOKENS]` with `MAX_TOKENS 50000` — a 400 KB stack array — and fails with:

```
error: source file exceeds maximum token limit (50000 tokens)
  help: split into multiple files using imports
```

Imported modules go through `aether_module.c`, which heap-allocates `MAX_MODULE_TOKENS 100000` ("to avoid a large fixed stack array; this also lets the cap grow without risking stack overflow"). So a module may be twice the size of the program that imports it, and the main file is the one place the limit still lives on the stack.

## Where it bit

aeb's fan-out orchestrator is a generated main file: one program calling every `.build.ae` node. aether-ui's 138-node tree generated 49,412 tokens (measured with the lexer); adding one node (aether-lang-dev/aether-ui#140) made it 49,770 + the root node and the whole build failed, with the error naming no file (it is aeb's own `target/_aeb/_orchestrator.ae`). aeb is fixing its side by emitting the per-node bookkeeping once (aether-lang-dev/aeb#15, 49,770 → 7,227 tokens), so this is not urgent for that consumer — but the "split into multiple files" help is not available to a generated single-entry program, and it is not available to anyone whose main file is simply large.

## Ask

Heap-allocate the main file's token array the way `aether_module.c` already does, and either share `MAX_MODULE_TOKENS` or make the cap large enough that it is a memory bound rather than a source-size one (a `Token*` is 8 bytes; 1,000,000 tokens is 8 MB). If a cap must stay, the error should print the file path — today it is the only diagnostic in the compiler that does not.

Measured with a 20-line tool linking `compiler/parser/lexer.c` (`lexer_init` + `next_token` until EOF) if a count is useful for the fix's test.

nicolas-maman:
Closes #2059.

## What was wrong

Four places lexed a whole source into a fixed-size array, each with its own cap and its own failure mode:

| site | cap | on overflow |
|---|---|---|
| main file (`aetherc.c`) | 50,000, on the stack | refused: "split into multiple files using imports" |
| imported module (`aether_module.c`) | 100,000 | silently truncated — tail declarations missing, `E0301` at the caller |
| `${...}` expression (`parser.c`) | 512 | silently truncated — `${x0 + … + x300}` printed the sum of the first 255 terms |
| LSP diagnostics (`aether_lsp.c`) | 4,096 | silently truncated — bogus syntax diagnostics on a long file |

The interpolation one is the worst of them: a wrong value with no diagnostic.

## Fix

One `lexer_tokenize(src, &count)` in the lexer returns a heap array that grows with the input and ends at the `TOKEN_EOF` (or the `TOKEN_ERROR` that stopped the scan); `free_tokens` releases it. All four sites use it; `MAX_TOKENS`, `MAX_MODULE_TOKENS` and `INTERP_MAX_TOKENS` are gone. The only failure left is out of memory, which the main-file site reports with the file name.

## Test

`tests/integration/source_size_unbounded` generates (into a temp dir, nothing large is committed) a 6,000-function main file, a 9,000-function imported module and a 601-token interpolation, and checks the exact result. Pre-fix each of the three fails in its own way; post-fix it prints `a=9002 b=13502 s=45150`.

Unit tests 398/398; sweep 1133/1134 on Windows, the one failure being `cache_dir_override`'s MSYS/native path spelling, which fails on `main` alone and touches no compiler path (filed separately).


nicolas-maman:
The macOS x86_64 failure was clang's 256-level bracket limit: codegen parenthesises every binary node, so the test's 300-term sum lowered to 300 nested brackets. That is a real limitation of the generated C on clang and is filed as #2071; the test here is reshaped to 150 `x * 1` terms — still 600 tokens, past the old 512 cap (pre-fix it still prints 8128 for the 11175 sum), and under the limit.

Issue #841, opened 2026-06-21
benchmark: replicate the concurrent-cache scaling study in Aether (actor-map vs sharded actors vs COW snapshot)


paul-hammant:
## Context

*"Shard your locks: benchmarking 6 Go cache designs"* (strebkov.dev, 2026-06) benchmarked six in-memory `string→string` cache designs across 1-8 cores and read/write mixes, and the rankings flipped by workload — with the "obvious" single-lock answer **negatively scaling** (0.66× at 8 cores). The methodology is worth borrowing to ground Aether's concurrency guidance in numbers rather than folklore (in the spirit of the #581 / #569 cross-reference rollups).

## Ask

Add an Aether benchmark replicating the study for the concurrency models Aether actually uses:

- **actor-owned map** — one actor owns the whole map (the "single mutex" baseline; expected to negatively scale)
- **sharded actors** — key routed by hash to one of N owner-actors (the lock-striping equivalent — see companion issue)
- **COW snapshot cell** — lock-free reads, copy-on-write writes (read-mostly winner — see companion issue)

Sweep:
- **cores** 1→8 (pin to physical cores — the article noted hybrid-CPU E-core/hyperthread spillover confounds unpinned sweeps; pinning left rankings unchanged but moved absolute numbers 10-25%)
- **read/write mix** read-only, read-heavy, balanced, write-heavy
- **key distribution** uniform + Zipfian (skew speeds up reads via cache locality but concentrates write contention — the article saw sharded drop to 0.82× under skew on the balanced mix)

## Output

Publish the throughput-vs-cores curves and a "what to actually use" table as concurrency guidance in the docs:
- default to **sharded actors** for general concurrent maps;
- **COW snapshot** for read-mostly config/routing/flags;
- single owner-actor only when contention is genuinely low.

Also a good forcing function to validate the sharded-actor helper and the COW cell once those land, and to find the shard-count knee for actors (the Go study's knee was ~256 *locks*; for actors it's likely core-count-scaled, not 256).

Related: the sharded-actor map issue, the COW snapshot issue, #736, #581, #569.

Issue #840, opened 2026-06-21
std: copy-on-write snapshot cell for read-mostly shared data (config / routing tables / feature flags)


paul-hammant:
## Context

Cross-reference from *"Shard your locks: benchmarking 6 Go cache designs"* (strebkov.dev, 2026-06). One design — **copy-on-write via `atomic.Pointer`** — owned the read-only and read-heavy workloads outright (87 Mops/s at 8 cores, fully **lock-free reads**, tracking or exceeding the ideal linear-scaling line). Its cost is all on the write side: every `Set` copies the whole map (the benchmark measured ~82 ms per write on a million-entry map), so it's strictly for **read-mostly-to-read-only** data.

## The gap in Aether

There's a recurring shape that fits neither "one actor owns it" nor "shard it": **read-mostly shared state that's written rarely and read constantly from many places** — config snapshots, routing/dispatch tables, feature flags, compiled rule sets. Aether has no first-class primitive for this. Routing every read through an owner-actor's mailbox (or any lock) is pure overhead when writes are near-zero.

## Ask

A **copy-on-write snapshot cell**: a typed holder over an immutable value where

- **reads are lock-free** — a plain atomic pointer load returns the current immutable snapshot; readers never block, never coordinate;
- **writes publish a new snapshot** — the writer builds a fresh immutable value and atomically swaps the pointer (compare-and-swap for read-modify-write); old snapshots stay valid for in-flight readers until released.

Sketch of the surface (names illustrative):

```
cell = snapshot.new(initial_value)      // holds an immutable T
v    = snapshot.load(cell)              // lock-free read, returns current T
snapshot.store(cell, new_value)         // publish a new snapshot
snapshot.update(cell) |old| -> derive_new(old)   // CAS-loop read-modify-write
```

This fits Aether's value/immutability leanings well: the snapshot is immutable by construction, so there's no aliasing hazard for readers.

## Guidance to ship with it

- **Use for:** config, routing tables, feature flags — anything read far more than written, where a write can afford to rebuild the whole value.
- **Do NOT use for:** write-heavy or large-value-frequent-write data — the whole-value copy per write is O(size) and dominates (the benchmark's 82 ms/write cautionary tale).
- It's the read-mostly complement to the sharded-actor map (companion issue): shard when writes are frequent and spread across keys; COW-snapshot when writes are rare.

## Notes

[...]

PR #2079, opened 2026-09-18
test(std.snapshot): concurrent copy-on-write — CAS loop never loses an update

paul-hammant:
## What

Adds a **concurrent** copy-on-write test for `std.snapshot` (issue #840): `tests/integration/snapshot_concurrent`.

## Why

`std.snapshot` had **no concurrent-writer coverage**:
- `snapshot_cow` (the existing regression test) exercises store/cas/load/reclaim but **single-threaded** — its own comment: *"single-threaded here, but it exercises the exact ownership moves a writer-actor makes."*
- The #841 concurrent-cache benchmark's COW design publishes with `store()` (unconditional overwrite), **not** a CAS loop.

So nothing in-tree proved the one property a real many-writer user of the cell depends on: **a `snapshot.cas` retry loop must never lose an update when many threads publish concurrently.**

This surfaced while building selaenium's distributed-Grid hub registry, which is `std.snapshot`'s **first concurrent-CAS-writer consumer**. The existing tests gave no confidence the concurrent path held — so this closes the gap in the primitive's home repo rather than leaving it proven only downstream.

## The test

Drives `std.worker` (a real thread pool) at one cell:

1. **500 concurrent CAS increments** → the cell must hold **exactly 500**. A broken `cas`, or a loop that doesn't retry the loser, loses updates and lands `< 500`.
2. **300 writers racing 300 concurrent lock-free readers** → writers still reach 300, and every acquire-load returns a real published value (**no torn reads**).

The `probe.ae` self-reports `snapshot_concurrent: N passing, 0 failing`; `test_snapshot_concurrent.sh` asserts on it and skips green without `ae`/threads. Passes on 0.685.


paul-hammant:
Verified beyond native Linux, since this is a threading/concurrency test — the two paths most likely to diverge from x86 Linux both pass:

| target | how | result |
|---|---|---|
| linux x86_64 | native | ✅ 500→500, 300+300 no torn reads |
| windows x86_64 | cross-build → Wine 9.0 (podman) | ✅ same |
| aarch64-linux | cross-build → qemu-aarch64 | ✅ same |

The Windows worker-pool path (CRITICAL_SECTION-based) and the ARM weaker-memory-ordering path both hold, so `std.snapshot`'s CAS acquire/release discipline and `std.worker`'s cross-platform pool are exercised on all three. (qemu-user isn't a perfect stand-in for real ARM-hardware concurrency — it can mask some ordering races — but it passes the ARM codegen path; the macOS-ARM64 CI leg runs it on real silicon.)

Issue #2083, opened 2026-09-18
Any foreign C thread → actor send is silently dropped (SPSC from_queues[MAX_CORES] illegitimate 2nd producer)


paul-hammant:
## Summary

A fire-and-forget `actor ! Msg` issued from a C thread that is **neither a scheduler thread nor the single main thread** is enqueued to a single-producer queue slot it does not own, and the message is **silently lost** — no crash, no diagnostic, the receiving actor's arm simply never runs.

**Severity:** high. `std.actors`' own module header (`std/actors/module.ae:26-30`) documents `whereis`-then-send from a handler as a first-class pattern, and that pattern is silently broken wherever the handler runs off-scheduler (the whole `std.http` worker pool, `std.worker` pools, embedding host threads, C-library callback threads).

**Discovered:** building selaenium's grid hub — a `std.http` handler sending to a registry actor (`a = actors.whereis("ctr"); a ! Inc {}`).

## Symptom

An HTTP handler on a `std.http` pool thread does `actors.whereis(name) ! Msg{}`. The handler completes normally (200, no crash), `whereis` returns a valid non-null ref, but the actor's receive arm **never executes**. The identical send from an **on-scheduler** context (inside another actor's receive) works.

## Minimal reproduction

One `.ae` file, `std.http + std.actors`, no external deps — send the SAME message to the SAME actor from two origins in one binary:

```
message Inc {}
actor Ctr { state n = 0
    receive { Inc() -> { println("[Ctr] Inc received"); n = n + 1 } } }

// (A) ON-SCHEDULER: from inside Srv's receive, before the blocking server loop
StartS(raw) -> { c = actors.whereis("ctr")  c ! Inc {}  http_server_start_raw(raw) }   // WORKS

// (B) POOL THREAD: from the http handler
inc_handler(req, res, ud) { c = actors.whereis("ctr")  c ! Inc {}  ... }               // DROPPED
```

`[Ctr] Inc received` fires exactly once (from A), never from B, across any number of POSTs. (Pool-thread stdout is block-buffered — run under `stdbuf -oL`; the handler DOES run, it's the send that's lost.)

Toolchain: ae 0.681.0, also present at HEAD.

## Root cause (file:line)

1. A `std.http` pool worker is a bare `pthread_create`'d thread (`std/net/aether_http_pool.c:121`). It never registers a scheduler identity, so its TLS `current_core_id` stays at the module default **-1** (`runtime/scheduler/multicore_scheduler.c:238`).
[...]

PR #2084, opened 2026-09-18
feat(std.sync): atomic i64 for pool-owned reclamation (#2082) + http actor-dispatch doc honesty

### LINKED ISSUE #2082: std.sync: expose a reclamation primitive (atomic int) to .ae for pool-owned COW structures
## Summary

`.ae` has no way to express **shared-state reclamation** for pool-owned concurrent data structures. `std.snapshot` gives an atomic pointer swap (publication), but nothing to safely free the *displaced* value once no reader can hold it. This proposes a small `std.sync` primitive to close that.

## Where it bites (concrete, shipped-with-a-caveat)

The selaenium distributed-grid hub registry is a pool-owned copy-on-write table (a `std.snapshot` cell, read lock-free by `std.http` handler threads, written via a `snapshot.cas` retry loop). It works and is correct for the lost-update property — but its `_retire()` **leaks the displaced table every write**, because a correct bounded retire-ring (push the displaced ptr, free the N-generations-old one once no reader can hold it) needs shared mutable state across concurrent CAS winners, and that ring itself needs a lock or an atomic. Released `.ae` exposes neither, so we shipped lock-free-COW with a documented per-generation leak. Fine for a rebuild-rarely registry; wrong if churn ever gets heavy.

This is **not** selaenium-specific — it's the exact shape the runtime's own reverse-proxy load balancer already implements **in C** (`std/http/proxy/aether_proxy_lb.c`: `atomic_fetch_add`, `atomic_load` + `pthread_mutex`). The capability exists one layer down; this lifts it into `.ae` so pool-owned COW structures don't have to drop to C for correct reclamation.

## What exists today (verified)

- No `std.sync`, no `std.mutex`, no atomic module. (`std.cas` is *content-addressed storage* — sha256 artifact keys — unrelated.)
- `std.snapshot` = atomic ptr swap only: publication, not reclamation.
- So the *only* `.ae` shared-state concurrency primitive is the snapshot cell.

## Proposal

Give `.ae` **one** of the following (either unblocks a correct retire ring / refcount):

**Option A — atomic int (recommended).** `fetch_add`, `load`, `compare_exchange` on an atomic int cell. Enough to build a refcount and a small lock-free retire ring. Smallest runtime surface; composes; and it's the same model `aether_proxy_lb.c` already uses — one locking model in the tree, not two. Keeps with actors-not-locks: this is low-level plumbing for the pool-owned-C-style seam the LB already occupies, not a general invitation to lock.

**Option B — `std.sync.Mutex`.** `lock`/`unlock` or a `with_lock` scope. Simpler mental model, but heavier runtime and invites the lock-ordering bugs the actor model exists to avoid.

Recommendation: **A**. It's the minimal thing that makes any pool-owned COW structure allocation-correct, and it matches existing C precedent rather than introducing a second concurrency idiom.

## Impact if it lands

- Grid registry's retire ring becomes allocation-correct (no per-generation leak); the "revisit if node churn gets heavy" caveat in `grid/registry.ae`'s header goes away.
- Any future pool-owned COW state (caches, routing tables, service registries) gets a correct-reclamation path in pure `.ae`.

## Not urgent

The current leak is acceptable for rebuild-rarely state. This is the one real Aether-level capability the distributed-grid build surfaced that can't be done from `.ae` today.

## Related (separate, already tracked elsewhere)

- `http_server_set_actor_handler` is half-wired (actor-dispatch doc describes an unusable `.ae` API).
- Foreign-C-thread → actor send is silently dropped (pool worker's `current_core_id == -1` enqueues to the SPSC `from_queues` slot as an illegitimate 2nd producer).

Neither is required for this; noted only because the same arc surfaced them.

paul-hammant:
Two related changes from the selaenium distributed-grid arc, which surfaced the Aether-level concurrency gaps.

## 1. `std.sync` — atomic i64 cell (closes #2082)

`.ae` had an atomic *pointer* swap (`std.snapshot`) but no atomic *integer*, so a reference count or a lock-free retire ring — what a pool-owned copy-on-write structure needs to reclaim its displaced values — could not be built in pure Aether. `std.snapshot` publishes; it does not reclaim. The grid registry currently ships a **documented per-generation leak** for exactly this reason.

New `std.sync`: one atomic 64-bit integer cell over a C11 `_Atomic(int64_t)`, mirroring `std.snapshot`'s structure (caps-accounted alloc, explicit acquire/release ordering, strong CAS).

```
c  = sync.atomic_new(0)
v  = sync.atomic_load(c)            // acquire
     sync.atomic_store(c, 5)        // release
n  = sync.atomic_add(c, 1)          // acq_rel, returns NEW value
n  = sync.atomic_sub(c, 1)          // acq_rel, returns NEW value
ok = sync.atomic_cas(c, exp, nw)    // strong CAS -> 1 / 0
     sync.atomic_free(c)
```

Design choices worth a look:
- **`add`/`sub` return the NEW (post-op) value**, unlike C's `fetch_add`, so the refcount-to-zero case is a plain `if sync.atomic_sub(rc, 1) == 0 { reclaim }` — no fetch-then-adjust across the language boundary.
- **acquire/release ordering** so an atomic gating other data carries publish/subscribe visibility (same discipline as `std.snapshot`'s pointer).
- **Deliberately minimal — not a mutex, not a general lock.** Low-level plumbing for pool-owned, C-style shared state (the shape `std/http/proxy/aether_proxy_lb.c` already uses in C), lifted into `.ae`. Not a license to replace the actor model with locks — that framing is written into the module header and README.

**Test** (`tests/integration/sync_atomic`): drives `std.worker` (a real thread pool) at one cell — (1) 500 concurrent `atomic_add` → exactly 500 (no lost increments); (2) a refcount released by 500 concurrent holders hits 0 **exactly once** (clean reclaim, no double-free / leak). 10/10 stable locally; valgrind clean (0 bytes in use at exit).

## 2. Doc honesty: actor-dispatch is C-internal, not a usable `.ae` API

`docs/http-server.md`'s "Per-connection actor dispatch" example presented a C-internal mechanism as a user-facing Aether API. On a release it does not compile: `unwrap_msg_http_connection` is not an exposed `.ae` extern (it appears only in the doc fragment), and the spawn/send/release fn-pointers `http_server_set_actor_handler` demands are not exposed either. This marks the section as C-internal-shape-only and documents the substrate that actually works for handler-shared state (pool-thread `std.snapshot` COW via the `ud` slot). `std.actors`' module header gains the matching caveat: a `!` send from an off-scheduler `std.http` handler is currently dropped (#2083).

## Related issues from this arc
- #2082 — the reclamation primitive (this PR)
- #2083 — foreign-C-thread → actor send silently dropped (runtime/scheduler; separate fix)

## Gates
`ae fmt`, stdlib-index, module-README, doc-block checks all green locally; clean valgrind on the smoke.

https://claude.ai/code/session_01YAS2Xtsem7MRFggdiEaakq


paul-hammant:
@nicolas-maman — flagging you for a design read on the `std.sync` half of this (part 1), since it adds a runtime C atomic primitive and touches the actors-not-locks line. The three questions are on #2082 (add/sub return-new-value vs C fetch semantics; the acquire/release choice; and whether an atomic-int-only, no-mutex primitive scoped as 'pool-owned-C-state plumbing' sits right with you at all). No rush, and happy to reshape the API before merge — the doc-honesty half (part 2) is independent if you'd rather that land separately.

paul-hammant:
**CI: green except one unrelated Windows flake.** All 25+ legs pass, including the ones that matter for a concurrency + C-atomics change:

- AddressSanitizer & LeakSanitizer (ubuntu + macOS) ✅
- Cooperative Scheduler (no pthreads) ✅ — `std.worker` inline path, test still 500→500
- Valgrind Memory Leak Check ✅, Memory Safety (Valgrind + ASan) ✅
- Linux GCC/Clang/Hardened, RISC-V, FreeBSD, Embedded ARM, macOS, **Windows MINGW64** ✅
- `integration_sync_atomic` (the new test) passes on every platform **including Windows/UCRT64**

The one red — `Build + Test (Windows / UCRT64)` — is **`integration_install_manifest`**, an `ae install` manifest shell test unrelated to this PR. My new test passed in that same run (`[PASS] integration_sync_atomic_test_sync_atomic`); the Makefile change touches only the `std/sync` source + objdir lines (nothing install-related); and UCRT64 was green on the two prior commits that ran it. It's a known-flaky Windows install test, not this change. A re-run should clear it.

paul-hammant:
**Caught up with main (0.689.0) and verified locally, incl. podman/wine.**

Rebased onto current main (absorbed 0.688.0, 0.689.0, and #2088's `modules = "."` — CHANGELOG-only conflicts, resolved). Then ran the full suite locally on two targets:

| target | build | result |
|---|---|---|
| linux x86_64 (native) | `make ae` | ✅ sync_atomic 3/3, 5/5 stable; retire-ring README example runs + valgrind-clean; null-cell aborts (SIGABRT) |
| windows x86_64 | cross-build → **Wine 9.0 (podman)** | ✅ sync_atomic 3/3, **5/5 stable**; null-cell aborts (exit 3, MSVCRT abort), fatal msg prints and the unreachable line never runs |

The two review fixes are exercised cross-platform: case 3 (concurrent CAS-loop → exactly 500) drives `atomic_cas` on the Windows CRITICAL_SECTION worker pool, and the null-cell fatal path aborts on both Linux and Windows. Stdlib-index and module-README doc gates green. Ready for another look.

Issue #1235, opened 2026-07-22
[regression] lib/ module edits still don't invalidate the build cache on v0.433.0 (stale binary, even with a different -o name)


nicolas-maman:
**Affected:** v0.433.0 (latest release, tag of 2026-07-22 - which already contains the #1025 fix merged in #1029 on 2026-07-06). **Platform:** Windows 11, toolchain built from source at the tag (MSYS2/WinLibs UCRT gcc 15.2).

## Summary

Editing a module under the default `lib/` directory still does **not** invalidate the build cache on v0.433.0: `ae build` reports `Built (cache hit)` and produces a binary containing the **old** module code. This is the same user-visible behavior #1025 described, reproduced on a release that includes the #1029 fix - so either the fix regressed between then and v0.433.0 or it doesn't cover this path.

Notably, the stale artifact is served **even when the output name changes** (`-o out1` then `-o out2`), so the cache key appears to consider the entry source but neither the imported module contents nor the requested output.

## Minimal reproduction

```sh
mkdir -p repro/lib/dep repro/src && cd repro

printf 'import lib.dep\n\nmain() {\n    dep.hello()\n}\n' > src/main.ae
printf 'import std.io\n\nhello() {\n    println("VERSION-1")\n}\n' > lib/dep/module.ae

ae build src/main.ae -o out1   # -> Built: out1.exe
./out1.exe                     # -> VERSION-1   (correct)

# change the module: different content AND different size
printf 'import std.io\n\nhello() {\n    println("VERSION-2-CHANGED-CONTENT")\n}\n' > lib/dep/module.ae

ae build src/main.ae -o out2   # -> Built (cache hit): out2.exe
./out2.exe                     # -> VERSION-1   (WRONG: stale module code)
```

Expected: the `lib/dep/module.ae` edit invalidates the cache and `out2` prints `VERSION-2-CHANGED-CONTENT`.

Actual: `Built (cache hit)` and `out2` prints `VERSION-1`. `touch`-ing the entry file does not help either.

## Workaround

Deleting everything under `~/.aether/cache/` makes the next `ae build` rebuild correctly - after clearing, the same build prints `VERSION-2-CHANGED-CONTENT` as expected.

## Real-world impact

We hit this in a live application: after editing a project `lib/` module, the rebuilt server binary silently served the previous module behavior (an HTTP endpoint kept returning the old response shape) while the build reported success - the worst kind of failure, because nothing looks wrong. Refs: #1025, #1029.

PR #2086, opened 2026-09-18
fs: native symlinks on Windows

nicolas-maman:
## What was wrong

`fs.symlink`, `fs.readlink` and `fs.is_symlink` were stubs on Windows that failed every call ("cannot create symlink" / "not a symlink" / 0), and the stdlib reference described them as POSIX-only. Separately, on every platform `fs.readlink` copied the raw extern's `strdup` result and dropped the original, leaking one allocation per call.

## Fix

- `fs_symlink_raw`: `CreateSymbolicLinkW`. A relative target is resolved against the link's own directory to decide whether it is a directory link (Windows records that); separators in the target are stored as backslashes, the only form the object manager resolves; `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` (Developer Mode) is tried first and the call repeated without it on a kernel that rejects the flag. Creating a link remains a privilege administrators hold and Developer Mode grants everyone.
- `fs_readlink_raw`: opens the link itself (`FILE_FLAG_OPEN_REPARSE_POINT`), reads its reparse data (`FSCTL_GET_REPARSE_POINT`) and returns the print name — the target as stored — falling back to the substitute name with its `\??\` prefix removed. A junction or any other reparse point is not a symlink. The symlink reparse layout is spelled locally since the SDK does not ship `ntifs.h`.
- `fs_is_symlink`: the directory entry's reparse tag (`FindFirstFileW`), no open.
- `fs_unlink_raw`: a link to a directory is removed with `RemoveDirectoryW`, which takes the link and leaves the target.
- `fs_readlink_raw` is declared `@heap` and `fs.readlink` hands the result straight back — no copy, no leak.
- `fs.symlink`'s doc states the Windows behaviour; the reference's platform table no longer lists symlinks as POSIX-only; a stale claim in `build-system.md` that the cache's lib-dir walk was POSIX-only is corrected (it has run on Windows since #1235, and `cache_libdir_invalidation` passes there).

## Verification

- `readlink`/`is_symlink`/`realpath` verified on real NTFS symbolic links on this machine (relative target returned verbatim, absolute target returned as stored, a junction reported as not a symlink); `symlink` creation could not be exercised here (no privilege, Developer Mode off — `ERROR_PRIVILEGE_NOT_HELD`), so `tests/regression/test_fs_symlink.ae` skips when the first link cannot be created and runs every case where it can, which is CI's administrator session.
- `test_fs_walk`, `test_fs_copy`, `test_fs_realpath` unchanged and passing; POSIX branch syntax-checked.


nicolas-maman:
The Wine leg: Wine answers `CreateSymbolicLinkW` by making a Unix symlink that carries no reparse tag, so the link was "created" but `is_symlink`, `readlink`, `realpath`-through-link and the listing kinds could not see it, and five tests that used to skip their symlink cases (the stub failed creation) now asserted on them. `fs_symlink_raw` now checks its own post-condition — the link the OS made must be one the OS reports — and otherwise takes it down and returns the same failure a missing privilege gives. Windows always passes the check (one `FindFirstFileW`); Wine goes back to the skip path.

PR #2099, opened 2026-09-18
feat(ae add): binary packages plus FreeBSD/arm64-Windows triples

paul-hammant:
Three related changes to `ae add`, building toward consuming published binaries as source-free packages (from the libphonenumber-ae / datastar-aether line).

## 1. Binary packages: a bare shared lib + a released `aether.toml`

A package that ships loose per-triple shared libraries (`lib<stem>-<tag>-<triple>.so`/`.dylib`/`.dll`) instead of a compilable source tree had no `ae add` path: `ae_try_release_asset` only fetched an **archive** (`<repo>-<tag>-<triple>.tar.gz`/`.zip`), so a bare `.so` was invisible and `ae add` fell through to `git clone` — which lands an uncompilable tree when the metadata tables are gitignored build artifacts.

New `ae_try_binary_package`, tried **before** the archive loop:
- fetch the release's `aether.toml`;
- if it carries `[package] binary = "<stem>"`, build `<stem>-<tag>-<triple>` + the host shared-lib ext, fetch + checksum-verify it, and install the **lib and the aether.toml** into the package dir;
- the manifest's `modules = "."` then puts the lib on the search path so the binary-import prepass fires.

**No fallback ladder** (the explicit design constraint): the `binary` key is the single signal.
- no `aether.toml` / no `binary` key → not a binary package, carry on to the archive path;
- `binary` key but no lib for this host → hard error naming the missing asset, **never a silent git clone**;
- mismatched checksum → fatal, same as the archive path.

Publisher gives the lib **stem** explicitly (`binary = "libphonenumber_ae"`), so `ae add` only appends what it can determine (tag, host triple, OS ext) — no hyphen/underscore/`lib`-prefix guessing.

## 2. FreeBSD + arm64-Windows host triples

`ae_host_triple()` returned `NULL` for FreeBSD and arm64 Windows, forcing those hosts down the git-clone path even when the publisher ships binaries for them. Added the `__FreeBSD__` (x86_64 + arm64) and `windows-arm64` cases.

## 3. CHANGELOG cleanup

Removed three stray `=======` lines (leftover conflict-marker middles from earlier release-fold merges).

## Verification

- New `tests/integration/ae_add_binary_package` — hermetic python http.server fake forge, all four properties: install-both + verified, missing-lib fatal, no-binary-key falls through, bad-checksum fatal. 3/3 stable.
- Existing `ae_add_release_artifact` (archive path) and `dep_resolution` still pass — archive/git paths intact.
- FreeBSD triple: per-platform preprocess selection verified + compiles under the real `zig cc` FreeBSD toolchain, both arches.

https://claude.ai/code/session_01YAS2Xtsem7MRFggdiEaakq


nicolas-maman:
Those three `=======` lines were mine — the script I used to re-seat a branch's `[current]` bullets after a release fold treated a bullet as running until the next heading, so the conflict-marker middle rode along inside the last paragraph. Fixed the script; a fourth one from the #2100 merge (after your branch was cut) is removed in #2102, in a hunk far enough from yours not to conflict. CI here is 28/28.

Issue #880, opened 2026-06-25
Allow type/keyword tokens (ptr, byte, func, state, ...) as value identifiers in C ports


paul-hammant:
## Summary

Several type/keyword tokens — `ptr`, `func`, `byte`, `state`, `match`, `after`, `union` — are reserved such that they can't be used as **value identifiers** (parameter/local names), even though they only have meaning in *type* or *statement-head* position. C ports hit this constantly because these are natural variable names (`ptr`, `func`, `byte`), forcing mechanical renames.

## Repro (aether 0.311)

```aether
f(ptr: int) -> int { return ptr + 1 }     // error: 'ptr' is a reserved keyword
f(func: int) -> int { return func + 1 }   // error: 'func' ...
f(byte: int) -> int { return byte + 1 }   // error
// also: state, match, after, union   ('fn' is fine)
```

## Impact (mquickjs port)

A C→Aether port inherits the C source's identifier names. `ptr`, `func`, `byte` are everywhere in C, so each becomes a rename: `ptr`→`ptr_`, `func`→`fn_val`, etc. This port carries **14 such workaround identifiers** (`ptr_` ×11, `fn_val` ×3) that exist purely to dodge the collision. Each is a small readability cost (`ptr_` reads as a typo) and a porting paper-cut repeated across every C codebase. `std.mem` is even named `std.mem` rather than the natural `std.ptr` *for this reason* (noted in its module doc).

## Why these are safe to allow in value position

The collision is between a **type/keyword** and a **value identifier** — two different syntactic positions:

- `ptr`, `byte` appear as *types* (`p: ptr`, `b: byte`). A `ptr` in value position (`return ptr`, `ptr + 1`, `ptr.field`) is unambiguously a value.
- `func` / `state` / `after` / `union` are statement/declaration heads or type-internal (`union` only inside `extern struct`). In an expression or as a bare param name they can't be confused with their keyword role.
- `match` is the one with real expression-position ambiguity (it heads a match expression), so it's the reasonable one to keep reserved.

Context-sensitive treatment — a bare identifier in value position is an identifier even if it spells a type keyword — would remove the friction for the unambiguous cases (`ptr`, `byte`, `func`, `state`, `after`, `union`) without touching the type grammar.

## Proposed change

[...]

PR #2117, opened 2026-09-19
compiler: seven fixes around reserved names, libc collisions and diagnostics (#2087, #2073, #2064)

### LINKED ISSUE #2064: A local named after a libm builtin (floor) breaks unrelated modules, with errors attributed to the wrong file
Seen with aetherc from Aether 0.651.0 (Windows, ae3d).

A program with a local variable named `floor`:

```aether
on_start(state: ptr, e: *Engine) {
    floor = loader.plane(2, 10.0)
    core.model_set_position(floor, 0.0, 0.0, 0.0)
}
```

fails type checking with three errors that point at a different file altogether:

```
error[E0200]: cannot cast ptr to int with `as`: a value cast converts between a distinct type and its base, or between numeric types
  --> examples/gltf_viewer.ae:127:18
  --> examples/gltf_viewer.ae:128:18
  --> examples/gltf_viewer.ae:129:18
```

`examples/gltf_viewer.ae` has nothing at those lines; the errors are in an imported module (`ae3d.noise`, which does `floor(x) as int`), where the builtin `floor` now resolves to the program's `*Model` variable. Editing the program does not move the line numbers, which is what makes it hard to find: I bisected the whole file before guessing the name.

Two asks:

1. A local should not shadow a builtin across module boundaries -- the module's `floor(x)` should stay libm's whatever a program names its variables -- or the compiler should refuse the shadowing with a message naming it.
2. An error inside an imported module should name the module's file and line, not the main program with the module's line numbers.

Renaming the variable (`ground`) fixes it.

### LINKED ISSUE #2073: A user function named "isolate" (and a local named "when") is silently ignored / rejected: reserved words should be an error at the definition
Aether 0.129 (the toolchain ae3d pins).

Defining a top-level function named `isolate` compiles without a diagnostic, and calls to it do nothing: the body never runs. In a 1,400-line tool I had

```
isolate(named: string) { ... sends a request ... }
...
isolate(mesh)
```

and the program behaved as if the call line were absent -- no error, no warning, no output from the body, the rest of the function proceeded. Renaming the function to `show_only` made it run. So `isolate` appears to be a word the language keeps (a builtin or keyword for something else), and a user definition of the same name is dropped instead of being refused.

Likewise a local named `when`:

```
    when = 0.0
```

fails with `E0100: Expected statement in block` pointing at the `=`, which at least fails, but names the wrong thing.

Expected: defining or assigning a reserved word is an error at the definition site that names the word as reserved (`'isolate' is a keyword`). Silently ignoring a call to a user function is the worst of the options -- it cost an hour to find in a port that was otherwise line-for-line.

Also seen, and worth the same message: `; ` after a closing `}` on the same line (`if x { a } ; b`) is `E0100: Expected statement in block` at the `;` rather than a note that a statement cannot follow a block on its line.

### LINKED ISSUE #2087: codegen: a closure parameter named `index` is renamed to `ae_index` in the signature but not in the body
Since somewhere after v0.651.0 (reproduced on v0.684.0; v0.651.0 emits `int index` and is fine), a closure parameter whose name collides with a libc symbol is mangled in the emitted signature but the body keeps the original name:

```aether
run(cb: fn(int) -> int) -> int { return cb(3) }

main() {
    f = |index: int| { return index + 1 }
    println("${run(f)}")
}
```

`aetherc repro.ae out.c` emits

```c
static int _closure_fn_0(_closure_env_0* _env, int ae_index) {
    return (index + 1);
```

so on Linux `index` resolves to libc's `index(const char *, int)` (`-Wint-conversion` warnings, wrong code), and on Windows it is `'index' undeclared (did you mean 'ae_index'?)`. A plain function parameter named `index` is left alone (`int plus(int index)`), so only the closure-parameter path renames. This broke ae3d's editor (four `callback |item: ptr, index: int, row: int|` closures over `ui.listbox`) the first time CI built it with a current toolchain; renamed there as a workaround.

nicolas-maman:
Closes #2087, closes #2073, closes #2064.

Seven fixes, each with a test, reviewed (four review findings folded in — see below). Full Windows sweep 1154/1154, unit tests 398/398, `fmt` gate clean, doc-block problem count unchanged (49 before and after; the reference edits are prose).

## 1. Closure parameter named like a libc symbol (#2087)

`|index: int|` emitted `_closure_fn_0(..., int ae_index)` with a body that read `index` — libc's `index(3)` on Linux, undeclared on Windows. The closure signature used `safe_c_name` (the *function*-name mangler, whose list carries libc symbols) while the body used `safe_value_name` (keywords only). The signature now uses the value mangler too: a keyword (`char`, `short`) is renamed in both places, a libc name in neither — a C parameter may shadow a libc function. `tests/regression/test_closure_param_libc_name.ae`.

## 2. A user function named after a builtin was silently ignored (#2073)

`isolate(named: string) { … }` compiled and every call did nothing: codegen lowers `isolate(x)` by name. The definition is now refused — `'isolate' is a builtin function and cannot be redefined … rename it (e.g. 'isolate_')` — for the arity codegen intercepts (a table mirroring the `strcmp(func_name, …)` dispatch, including each name's arity condition). A definition outside that arity (a two-argument `isolate`, a zero-argument `select`) stays the plain function it always was; builtins codegen does not intercept (`each`, `map`, `filter`) or that yield to a user definition (`atoi`) are unaffected.

## 3. `when = 0.0` (#2073)

Was "Expected statement in block" pointing at the `=`. A statement keyword followed by an assignment operator is now refused at the keyword with the rename named. #880's value-identifier tokens (`func`, `state`, `after`, `ptr`, `byte`) are excluded, and recovery stops at a `}` on the same line so `if x { message = 2 }` reports exactly one error.

## 4. `if x { a } ; b` (#2073)

Was refused at the `;`. Every simple statement already consumes an optional `;`; a block-ending statement did not. A `;` at statement position is accepted as the terminator of whatever came before.

## 5. A closure passed where the parameter is a typed C function pointer

`run(f)` with `run(cb: fn(int) -> int)` and a closure reached the C compiler (`expected 'int (*)(int)' but argument is of type '_AeClosure'`). Refused at the call — a closure carries an environment, a C function pointer has none — naming the two spellings that work (a bare `fn` parameter with `call(cb, …)`, or a named function `as fn(...)`). A named function still passes, with or without the cast.

## 6. Module diagnostics named the main file (#2064, second ask)

Modules are merged before checking, so an error at `noise/module.ae:4:17` was reported as `main.ae:4:17` — a location that exists and is wrong. Every diagnostic now carries the file of the top-level definition being checked; the reporter already reads a module file for its snippet.

## 7. A user function named after a libc symbol, used as a value (#2064)

`read(thing: *Thing) -> int` is defined as `ae_read` and a direct call reached it, but `read as fn(*Thing) -> int` and the adapter behind `g = read` emitted the bare name: libc `read`'s `-1` at runtime on Windows, a `-Wint-conversion` warning and wrong code on Linux. The cast operand is marked by the typechecker and spelled with the definition's name (after the `@c_callback` lookup, whose bound symbol wins for those functions); the adapter calls the definition's spelling. `tests/regression/test_fn_libc_name_as_value.ae`.

The first ask of #2064 (a local named `floor` shadowing a module's extern) no longer reproduces on `main`.

## Review

`/code-review high` on the diff found four problems, all fixed and pinned in `tests/integration/reserved_name_diagnostics`: the keyword check also caught #880's value-identifier tokens; the builtin refusal was over-broad (every node-less builtin, including `each`/`atoi`); the `fn_addr` spelling ran before the `@c_callback` symbol lookup; error recovery ate a `}` on the same line. The clean sweep then caught a fifth (a zero-argument user `select`, which codegen never intercepts), which is what made the refusal arity-aware.

## Docs

`docs/language-reference.md`: `;` after any statement; keywords are not names; builtin names cannot be redefined while libc-named functions are fine in every position; a closure cannot go to a typed fn-pointer parameter.


nicolas-maman:
Both red legs were one thing: `make check-contrib-modules` type-checks each module file on its own, where `contrib/sqlite`'s `free(rs: ptr)` is a top-level definition named `free` — and the new builtin-redefinition rule refused it. A module's functions live in the module's namespace (`sqlite.free(rs)` is never a call by bare name; in a merged program the definition already carries its prefix), so the rule now applies to programs only, keyed on the file's `exports(...)`. `check-contrib-modules` is 9/9 locally and the case is pinned in the test. The sweep on that leg was 1154/1154 before the step failed.

PR #2121, opened 2026-09-20
fix: array literal element join, inferred-int narrowing, uint32 printing and identifier-typed array declarations

### LINKED ISSUE #2119: typeck/codegen: an array literal of float expressions (0.18 * math.PI) is typed as int and truncates every element
An array literal whose elements are float *expressions* involving a module constant comes out as an int array with the values truncated. Literal floats, a subtraction of literals, and a bare module const in the same position are fine.

Reproducer (ae 0.651.0, Windows; `math.PI` is a `const PI = 3.14159265359` in an imported module):

```
import std.string
import aephysics.math

main() {
    rolls = [ 0.18 * math.PI, 0.25 * math.PI, 0.32 * math.PI ]
    x = 0.18 * math.PI
    println("rolls ${rolls[0]} ${rolls[1]} ${rolls[2]} direct ${x}")
    yaws = [ 0.0 - 0.35, 0.0 - 0.15, 0.0, 0.15, 0.35 ]
    println("yaws ${yaws[0]} ${yaws[1]} ${yaws[4]}")
}
```

prints

```
rolls 0 0 1 direct 0.565487
yaws -0.35 -0.15 0.35
```

Expected `rolls 0.565487 0.785398 1.005310`. The same elements assigned one at a time are right, so it is the literal's element type: it looks as if the element type is taken from something other than the expression's type (int?), and each element is converted through it. A `const ROOT2 = 1.41421356` used bare as an element (`[ 1.43, ROOT2, 1.40 ]`) is fine, so it is the multiplication that trips it, not the const.

No error or warning is emitted, which is what made this costly: a test in aether-lang-dev/aephysics ran a sweep with all its angles at 0 or 1 and only a count assertion caught it.

nicolas-maman:
Closes #2119.

Eight fixes around numeric typing of array literals and the unsigned widths, each with a regression test, a CHANGELOG bullet under `[current]` and reference updates.

## Fixes

1. **Array literal of float expressions typed as int (#2119).** `[0.18 * math.PI, 0.25 * math.PI]` was typed by the early inference pass from its first element before the module constant resolved (unknown → int), so every element truncated silently. The element type is now the numeric join of all elements once typed (float if any is float, `long` if any needs it). Three sites had to agree: the literal, the inferred declaration that adopts it, and the index expression that was stamped with the stale element type and reached `printf` as `%d` on a double.
2. **A bare local bound to an element of such an array followed the stale stamp.** `xs = [1, 2.5, 3]; k = xs[1]` emitted `int k` (printed 2); `z = longs[0]` emitted `int z` (printed `-294967296`). An inferred declaration whose early `int` stamp disagrees with a now-typed float/64-bit initializer adopts the initializer's type.
3. **A `uint32`/`uint16`/`uint8`-led literal keeps its element kind.** Only a 64-bit element widens an integer join, so `uint32 x = ...; arr = [x, 0]` stays `uint32_t arr[2]`.
4. **A float assigned to an inferred `int` variable printed garbage.** `r = 0; r = 2.5` kept `int r` while the checker typed the name as float, so `${r}` went through `printf("%g", int)`. It is now the #698 narrowing error, with the float wording (`float r = ...` to keep the fraction, `int r = ...` to narrow explicitly).
5. **A float assigned to an element of an inferred int array truncated silently.** `xs = [1, 2]; xs[0] = 7.5` is refused the same way. Only literal-initialised int arrays are marked inferred, so an `int[]` returned by a call is not told its type "was inferred from the array literal". Annotated slots (`int k = 4.5`, `int[2] xs`) still narrow on purpose.
6. **A nested array literal emitted invalid C.** `[[1.0, 2.0], [3.0, 4.0]]` lowered to `int grid[2] = {{...}, {...}}` and failed in the C compiler. Arrays are one-dimensional; the inner literal is refused with the flat-array and list-of-arrays spellings named.
7. **A `uint32` printed as a negative int.** Every print path (`print(x)`/`println(x)` statement and expression forms, `print("%d", x)`, `${x}` interpolation, `_aether_interp`) chose `%d`, so `uint32 x = 4000000000` printed `-294967296`. `uint32_t` is `unsigned int` on every supported target; `%u` is its conversion.
8. **`uint32[4] xs = [...]` did not parse as a declaration.** `int[4] xs` reaches the typed-declaration path through the keyword; `uint8`/`uint16`/`uint32` and struct names are identifiers, and only `IDENT IDENT` was recognised, so `uint32[4] xs` was an index into an undefined `uint32` and `Pair[2] ps` emitted invalid C. The shape `IDENT [N] IDENT` on one line is now a declaration; the element type is pinned in the C (`uint16_t hs[3]`, `Pair ps[2]`).

## Review

`/code-review high` on the first cut found three regressions (2, 3 and the wording in 5 above); each is fixed and pinned in the tests. Every "before" behaviour above was reproduced against an origin/main build in a scratch worktree, not assumed.

## Tests

- `tests/integration/array_literal_element_type` — the #2119 program (module constant, call, mixed, all-int, long-led literals, derived locals, narrow-led literals), the three refusals, the annotated forms that still narrow, and the `int[]`-from-call wording.
- `tests/integration/uint32_print` — every print path for a uint32 past 2^31, plus uint16/byte.
- `tests/integration/ident_typed_array_decl` — uint32/uint16/uint8/struct local arrays, their C element types, and a bare index-expression statement that must keep parsing as one.

Local gates on Windows (MSYS2 UCRT64): unit tests pass, `check-contrib-modules` 9 ok / 0 failed, `check_changelog_fold.sh` clean, doc-block check with zero problems in `language-reference.md`, full `test-ae` sweep 1158 passed / 0 failed.

## Docs

`docs/language-reference.md`: primitive-types table gains the unsigned widths; array declarations document identifier-spelled element types and `uint32` printing; the array-literal join rule and the `float → int` conversion row (explicit slots truncate, inferred slots error).


nicolas-maman:
CI was red on every lane for one cause: `make check-docs` refused the stdlib reference's math example, which rebound `c` from `math.clamp_int(...)` to `math.cos(0.5)` — the exact silent truncation fix 4 now diagnoses (the example was computing `c = 0`). The example binds the cosine to `co` now; the full `check_doc_blocks.py` run is clean locally (185 blocks compile, 61 run and match).

PR #2128, opened 2026-09-20
[ask] gcc 16 -O2 miscompiles aetherc-emitted C: IPA bit-CP folds a live int param to 0

paul-hammant:
An `ask` report (doc only — no toolchain code change in this PR) from the selaenium line. Filed by se-co (CachyOS, the failing box) + confirmed by the ChromeOS box; opened here because the failing box has no `gh` CLI.

## Summary

On **gcc 16.2.1**, `-O2` (the flags `ae build`/`aetherc` use — `-O2 -pipe -Wformat -fwrapv`) miscompiles aetherc-emitted C: `-fipa-cp` + `-fipa-bit-cp` **fold a live `int` parameter to the constant `0`**. In the Grid hub's `registry__row`, the 5th SysV arg (`inuse`, `%r8d`) is dropped and `string_from_int` is fed `xorl %edi,%edi` instead of the real value. Any aether program with a function whose parameter is a literal at one call site and a runtime value at another is exposed — not just Grid.

## The one-instruction diff, at compile time (no link, no run, no libaether)

    aetherc --with=net,os,fs --lib grid --lib selenium_core --lib selenium_core/drivermgr grid/hub.ae /tmp/hub_exe.c
    gcc -E -fwrapv $(ae cflags --cflags) /tmp/hub_exe.c -o /tmp/hub_exe.i
    gcc -O2 -S /tmp/hub_exe.i -o /tmp/bad.s     # grep registry__row

    gcc 16 -O2:               xorl %edi,%edi   # from_int(inuse) -- WRONG, folded to 0
    gcc 16 -O2 -fno-ipa-bit-cp / gcc 12 -O2:  movl %r12d,%edi   # the real 5th arg

Necessary + sufficient trigger proven both directions: removing either `-fipa-cp` or `-fipa-bit-cp` from `-O2` fixes it; adding both to a clean `-O1` breaks it. ASAN and UBSAN are both silent (no memory error, no UB — just wrong code). Cross-compiler control: gcc 12.2.0 emits `movl %r12d,%edi` from the byte-identical `.i`; gcc 16.2.1 emits `xorl`.

## The ask (primary: this repo)

Add **`-fno-ipa-bit-cp`** to the release C flags `ae build`/`aetherc` pass to gcc, and to the flags used to build the toolchain itself (an `ae`/`aetherc`/aeb-`tools` binary built by gcc 16 `-O2` is exposed too). `-fno-ipa-cp` also works but disables more; `-fno-ipa-bit-cp` is the minimal disable. Gating on gcc ≥ 16 is fine; unconditional is safer (upper bound unknown). The end-to-end fix is verified: `AE_CC="gcc -fno-ipa-bit-cp" aeb grid/.build.ae` → correct, on a cold `~/.aether/cache`.

Side note for whoever verifies: neither `~/.aether/cache` nor aeb's cache keys on the C compiler/flags, so a rebuild silently serves old-compiler objects — clear `~/.aether/cache` first or you measure the cache.

Full report in the added `asks/` doc. selaenium's grid code is correct and unchanged.

https://claude.ai/code/session_01N1Vjg3yn7uc7uMXNmyJz1e


nicolas-maman:
Thanks for the precise write-up. I can't reproduce it here: `grid/registry.ae` isn't in the pushed selaenium history (main has only `hub.ae`/`hubcore.ae`, and `7fad6ee` isn't a ref on origin), so there is no `.i` on this side to put through gcc 16.2.0, which is what MSYS2 UCRT64 has.

Before we add a flag: an IPA-CP + bit-CP fold of a parameter to `0` when one call site passes a literal `0` and the other "genuinely varies" is exactly what gcc is *allowed* to do when the varying value is **indeterminate on some path** — reading an uninitialized automatic is undefined, not "some garbage value", so the bit lattice legitimately meets `0` with "anything" and yields `0`. ASAN and UBSAN are both silent on that (only MSan sees it), "every probe fixes it" is the signature, and gcc 12 happening to keep the register is not evidence either way. aetherc emitted exactly that shape: a local first bound inside an `if`/loop body and used after it is hoisted to function scope as a bare `int inuse;` (`hoist_if_branch_vars` / `hoist_loop_vars` in `codegen_stmt.c`), so any path that skips the body reaches the use with an indeterminate value. In #2131 every hoisted declaration now carries its kind's zero initializer (`= 0` / `= NULL` / `= {0}`), which makes such a read a defined 0 — regardless of which compiler does the constant propagation, and without turning an optimization off for everyone.

I can't claim that is your chain without seeing it. Two things would settle it:

1. The `.i` (or the emitted `hub_exe.c`) from the failing box, attached here or mailed — one `gcc -O2 -S` + grep on my gcc 16.2.0 will say whether the fold survives the zero-initialized hoists.
2. The `registry.ae` / handler source for the `inuse` chain, to see whether any binding along `_int_field → registry_report_inuse → registry__set_inuse → registry__drop_and_add → registry__row` is a hoisted branch-local.

If the fold survives with every local defined, it's a real gcc 16 bug and `-fno-ipa-bit-cp` is the right response (gated on the compiler version, with the reduced `.i` filed upstream). If it doesn't, the flag would have papered over UB in the emitted C, and the fix is the one already in #2131. The `asks/` doc is a good record either way; I'd rather it name the outcome once we have the input.

paul-hammant:
[Paul's Claude `se` writing on behalf of Paul's Claude `se-co`]: @Nic — you were right, thank you. I've corrected the mechanism in the PR title and body (the original "folds the literal 0 from one call site" framing was wrong).

Your objection was the correct one to raise: an indeterminate read would license the fold, ASAN/UBSAN silence + "every probe fixes it" fit that signature, and aetherc does hoist branch-locals as bare decls (aether#2131). se-co tested it on the failing box and ruled it out three ways:

1. `-ftrivial-auto-var-init=zero` (and `=pattern`) — defines *every* automatic in the TU, a strict superset of aether#2131 — and the fold survives. `-Wmaybe-uninitialized -Wuninitialized` over the whole TU: 0.
2. Zero bare-uninitialised scalar decls on the `inuse` chain (handler → report_inuse → set_inuse → drop_and_add → row); each binding is initialized at its declaration and total on every path. The hoisted `int nl; int max; …` shape is in the file (registry__pick/_find_row/_drop_row) but none feeds param #4.
3. `-fdump-ipa-cp-details` shows the value lattice stayed **VARIABLE** — not a constant fold. It's the **known-bits mask**: `param[4] Bits: mask = 0xf..f00000000` (low 32 known-zero) while the same param's **irange** correctly carries `MASK 0x7fffffff` (the `if (inuse<0) inuse=0` clamp → 31 unknown bits). The Bits lattice dropped exactly those unknown bits. Correct mask `0xf..f7fffffff` appears 11× elsewhere in the dump, so it's representable; control `registry__f`'s `int i` (0..5) gets the right `0xf..f00000007`.

So: bit-CP's known-bits meet drops the unknown bits for a clamped non-negative int param, not IPA-CP concluding it's constant. Same wrong `xorl`, different (correct) mechanism — and it matters for a gcc filing since it points at the bits meet, not constant propagation.

Full analysis + the ruled-out-UB grounds are in the selaenium `asks/` doc (recorded in selaenium `3997957`). aether#2131 is a good change on its own merits — just orthogonal to this. The remedy (`-fno-ipa-bit-cp`) is unchanged. se-co's user has the self-contained `.i` + the `cp.dump` to attach.

(For the record on process: I opened this PR from se-co's branch because their box has no gh CLI — the analysis is theirs; I'm relaying the correction here since I hold the edit access.)

paul-hammant:
@Nic — stronger result: se-co retested against **#2131 itself on the released 0.701.0**, not a proxy. The earlier three grounds were run on 0.699.0, which predates your fix — a real gap, now closed. The fold survives your fix.

**#2131 is present and working in 0.701.0's emitted C** — worth saying plainly, it's a good fix and does what it says:

    0.699.0:  int nl;      int line_end;      int max;
    0.701.0:  int nl = 0;  int line_end = 0;  int max = 0;

Bare uninitialized scalar decls across the hub TU: **436 → 144**, and all 144 that remain are **struct fields** (`struct LocalTime { int year; ... }`), not function locals. **Zero uninitialized function locals remain in the TU.** #2131 is complete for its purpose on this input.

The fold survives it, asm and runtime:

    aetherc 0.701.0 + gcc 16.2.1 -O2                              -> FOLDED to 0   "inuse":0
    aetherc 0.701.0 + gcc 16.2.1 -O2 -fno-ipa-bit-cp              -> real arg      "inuse":3
    aetherc 0.701.0 + gcc 16.2.1 -O2 -ftrivial-auto-var-init=zero -> FOLDED to 0
    aetherc 0.699.0 + gcc 16.2.1 -O2                              -> FOLDED to 0   (unchanged)

(Runtime legs: hand-link of hub_701.c against 0.701.0's own libaether, POST maxSessions:7/inuse:3, GET /se/grid/status. Chain functions under 0.701.0 carry zero bare uninit locals.)

So your own criterion — *"if the fold survives with every local defined, it's a real gcc 16 bug and -fno-ipa-bit-cp is the right response"* — has now been met literally, using your fix, on the released toolchain. The mechanism (bit-CP known-bits mask dropping the unknown bits of the clamped `inuse` param; value lattice VARIABLE, irange MASK 0x7fffffff vs Bits mask 0xf..f00000000) stands, and #2131 is orthogonal — a good change, just not this bug.

**Honest caveat:** aether 8f1c041b and 127a348d (further hoisted-local fixes) land *after* 0.701.0 and are not in this test — worth a re-run on 0.702.0, which se-co has queued.

The .i to attach is the new one with your fix compiled in: **hub_701.i.gz**, sha256 `d435aded307f684400932c570b2dc9e9fb9142300060c55d6800fffddb59eb43` (supersedes the 0.699.0 .i). se-co's user has it. Recorded in selaenium `c954a6c`.

(Minor side note, last: the earlier premise that `grid/registry.ae` isn't in the pushed history — it is: `git ls-tree origin/main grid/` lists it, and `7fad6ee` is a commit and an ancestor of `origin/main`, not a branch ref, so `git show 7fad6ee:grid/registry.ae` or a checkout by SHA both work. Doesn't affect the substance — you were right to push on the mechanism.)

paul-hammant:
Caveat closed: **8f1c041b + 127a348d are in 0.702.0 and the fold survives both, as does 0.706.0** (newest published).

se-co re-ran the same hub.ae / same method on gcc 16.2.1:

| toolchain | `-O2` | `-O2 -fno-ipa-bit-cp` | `-O2 -ftrivial-auto-var-init=zero` | runtime `-O2` |
|---|---|---|---|---|
| 0.701.0 | FOLDED | ok | FOLDED | `"inuse":0` |
| 0.702.0 | FOLDED | ok | FOLDED | `"inuse":0` |
| 0.706.0 | FOLDED | ok | FOLDED | `"inuse":0` |

Bare uninitialized scalar decls hold at 144 across 0.701/0.702/0.706 — all struct fields, zero function locals, zero in all five chain functions. So the hoisted-local line of fixes (#2131 and its follow-ups) is **complete and does what it set out to do** — and it's **orthogonal** to this. The `-fno-ipa-bit-cp` remedy is unchanged; the bit-CP known-bits-mask mechanism stands.

One honest note, because it's this investigation's own kind of failure: se-co's *first* scripted 0.702.0 run reported the asm as clean while the runtime still returned `"inuse":0` — a blind detector, not a fix. (`xorl[ \t]+%edi, %edi` doesn't match a tab under GNU grep — in an ERE bracket expr `\t` is the set {space, backslash, t}; it worked interactively via a ugrep shim, failed silently under `#!/bin/bash`.) It was caught precisely because the runtime leg disagreed with the asm leg — the argument for keeping two independent measurements even when one seems sufficient. Every interactive result already posted here was produced by the working path and is unaffected; the detector now self-tests (must match a known-folded line AND reject a known-good one) before printing any verdict. Recorded in selaenium `4675c79`.

So no re-run outstanding: the fold persists across every toolchain that carries the hoisted-local fixes.

PR #2131, opened 2026-09-21
fix: foreign-thread actor sends arrive; spawn_* user calls; struct clashes across modules; hoisted locals zero-initialized

### LINKED ISSUE #2083: Any foreign C thread → actor send is silently dropped (SPSC from_queues[MAX_CORES] illegitimate 2nd producer)

### LINKED ISSUE #2126: A call to a user function named spawn_* is emitted with only its first argument
## Reproduce

```aether
import std.string
f(x: int, n: int, p: ptr) { println("${x} ${n}") }
spawn_once(x: int, n: int, p: ptr) { println("${x} ${n}") }
main() {
    block = null as ptr
    f(1, 2, block)
    spawn_once(1, 2, block)
}
```

`aetherc repro.ae repro.c` emits

```c
f(1, 2, block);
spawn_once(1);
```

and gcc then fails with `too few arguments to function 'spawn_once'; expected 3, have 1`. No diagnostic from the typechecker. Renaming the function (`one_spawn`) emits the full call, so the `spawn` prefix of the identifier is being taken for the actor `spawn` form somewhere in the parser or the call lowering, and the remaining arguments are dropped silently.

Found porting Box3D's test_determinism.c (`QuerySpawnOnce` → `spawn_once`) in aether-lang-dev/aephysics. Expected: an identifier that merely begins with `spawn` is an ordinary function; if `spawn` is reserved as a prefix, the compiler should say so.

### LINKED ISSUE #2129: Struct names are one namespace across modules: two modules defining the same struct name silently merge
## What happens

Two modules that each define a struct of the same name compile without a diagnostic, and one definition wins for the whole program. Reproduced integrating aether-lang-dev/aephysics into nicolas-maman/ae3d: both define `Quat` (`{v: Vec3, s: float}` in the physics library, the reference's b3Quat; `{x, y, z, w}` in the engine), `Plane`, `Bone`, `Buffer`, `World`. A program importing `ae3d.core` and `aephysics.math` emitted one `Quat` (`{x, y, z, w}`) and the physics library's `q.v` / `q.s` then failed in the C compile — the first the author hears of it. With identical definitions (both engines' `Vec3`) nothing fails and the merge is silent.

`exports (...)` scopes functions and constants per module; struct (and actor) names are not scoped: docs/module-system-design.md lists "Exporting structs and actors from modules" under Future, and nothing documents the merge.

## What it costs

Library authors must pick globally unique struct names by convention (the C `b3` prefix, which Aether's modules were meant to make unnecessary), and integrating two independently written libraries means renaming in one of them: aephysics#30 rewrote its quaternion to the engine's layout and ae3d renamed `World`, `Plane`, `Buffer` to `VoxelWorld`, `FrustumPlane`, `GltfBuffer`.

## Asks

1. Diagnose the clash: two different definitions of one struct name across the import closure is an error naming both modules; two identical definitions could be allowed (one type) or warned.
2. Scope struct names per module, like functions: `math.Quat` and `core.Quat` distinct, with `import` bringing the names in and clashes reported at the import.

nicolas-maman:
Closes #2083, closes #2126, closes #2129.

Five fixes taken from the open issues, by severity: silent message loss from foreign threads, silent argument dropping, silent struct merging across modules, and the undefined read behind the gcc 16 report. Each has a regression test, a CHANGELOG bullet under `[current]` and doc updates.

## Fixes

1. **A `!` send from a thread that is not a scheduler core was silently lost (#2083).** Every non-scheduler producer — the main thread, a `std.http` pool worker, a `std.worker` thread, a C library's callback thread — enqueues into the target core's `from_queues[MAX_CORES]` channel, whose SPSC producer side tolerates one writer; two such threads writing it at once raced on `tail`. The channel now serializes its non-scheduler producers under a per-core lock (`foreign_lock`); scheduler-thread sends are untouched, and the batch fan-out path takes the same lock. Paul's root-cause analysis in the issue was exact.
2. **…and with a single actor it corrupted the heap instead.** One actor puts the runtime in main-thread mode — the thread running `main()` steps the actor inline at each send — and a foreign thread's send stepped the same actor at the same time. A thread that is not the main-mode thread now leaves the mode (as spawning a second actor does) and sends through the scheduler; the inline step holds the actor's `step_lock` from before the mailbox write until the step returns, with owner tracking so a handler's self-send on the same thread does not spin on its own lock, and a sync send that finds the mode gone hands the message back to the heap-copying path. Verified both ways on the unpatched runtime: the one-actor program crashed (`0xC0000374`), the two-actor program hung in the corrupted channel.
3. **A call to a user function named `spawn_*` was emitted with only its first argument (#2126).** Codegen took any `spawn_` prefix for the generated actor spawner. Only the spawner is treated that way now; a user function or extern that starts with `spawn_` is an ordinary call.
4. **Two modules defining one struct name differently merged silently (#2129, ask 1).** The clash is reported at the second definition — `struct 'Quat' is defined differently in two modules: {v: Vec3, s: float} in lib/phys/module.ae and {x: float, y: float, z: float, w: float} in lib/geo/module.ae` — and stops the compile (a new gate after the merge; the pruner could otherwise drop the losing side and let the program build). Identical definitions still share one type. The scan this enabled found two clashes inside the standard library: `std.cryptography.asn1`'s internal `Reader` vs `std.cbor`'s, and `std.message`'s internal `Parser` vs `std.jsonpath`'s — renamed to `Asn1Reader` / `MessageParser` (module-internal, nothing exported changes). Ask 2 (per-module struct scoping) is a design change and stays open.
5. **Hoisted locals are zero-initialized.** A local first bound inside a branch or loop body and used after it is hoisted to function scope as a bare `int x;`; on a path that skips the body its value is indeterminate, and reading an indeterminate value is undefined — a C compiler's interprocedural constant propagation may treat it as whatever the other call sites pass, with no sanitizer able to say so. That is the shape of #2128's gcc 16 `-O2` report; I could not reproduce that report (its `registry.ae` is not in the pushed selaenium history) and do not claim this is its cause — the comment on #2128 says what would settle it. Every hoisted declaration now carries its kind's zero initializer (`= 0` / `= NULL` / `= {0}`, chosen per kind so clang's `-Wbraced-scalar-init` stays quiet).

## Review

`/code-review high` found six real problems in the first cut and all are fixed and pinned: three re-entrancy deadlocks of the new step lock (a handler's self-send on the same thread — single-int, repeated, and from an inline step), the mailbox-write-before-lock window that let a freshly started scheduler thread free a stack payload, the missing compile gate after the merge diagnostic, and three type kinds the zero-initializer spelled as `{0}`.

## Tests

- `tests/integration/foreign_thread_send` — two `std.worker` threads and the main thread each send 20 000 messages to one actor concurrently, in both runtime modes (single actor / second actor spawned first); the count must be exact.
- `tests/integration/spawn_prefixed_function`, `tests/integration/struct_clash_across_modules` (the clash message, the compile gate incl. unused imports, program-vs-module, identical definitions kept), `tests/integration/hoisted_local_zero_init`.

Local gates on Windows (MSYS2 UCRT64, gcc 16.2.0), from a clean build: numbers in the last comment.

## Docs

`docs/http-server.md` "Handler → actor from a release" and `std/actors/module.ae`'s header no longer describe the drop as a caveat; they describe the ingress channel and keep the read-side guidance (`std.snapshot`).


nicolas-maman:
Local gates on Windows (MSYS2 UCRT64, gcc 16.2.0), from a clean build: unit tests pass, `check-contrib-modules` 9 ok / 0 failed, `check_changelog_fold.sh` clean, `ae fmt --check std` clean, `check_doc_blocks.py` 185 blocks compile / 61 run and match, `check_doc_examples.py` / `check_stdlib_index.py` / `check_module_readmes.py` clean, full `test-ae` sweep **1163 passed / 0 failed**.

